PR #46, opened 2015-04-14
IDEX-2368: Fixed problem when parent of repository received from list is empty

vinokurig:


vinokurig:
merged manually by squash merge

PR #62, opened 2015-04-17
IDEX-2286: configure RAM and timeout

ashumilova:
@vparfonov :
1. Fix properties alignment
2. Fix default project switcher on panel
3. Add popup on user configuring runner with RAM more then available
4. Add runner timeout configuration to user preferences
5. Set timeout on runners panel due to workspace attributes or defualt (4h)
6. Fix test due to changes


vparfonov:
ok

PR #66, opened 2015-04-17
IDEX-2404 Copy scope environment from System scope has wrong type

benoitf:
@svor what do you think ?


benoitf:
@svor here is the updated based on comments

svor:
@benoitf ok

vparfonov:
Ok

PR #71, opened 2015-04-20
IDEX-2402: Fixed bug related to incorrect environment name. Added environment name validation.

dimasnurenko:
@svor @vparfonov


vparfonov:
ok, only add comet here https://github.com/codenvy/che-plugins/pull/71/files#diff-fc916a717a688f3a1291bba4cc52b457R70 why we fix it in this way

PR #99, opened 2015-05-05
Idex 2121

vitaliy-guliy:


skabashnyuk:
why DTO's doesn't have setters?

PR #109, opened 2015-05-15
IDEX-2460: Fixed perform Git push operation after using github - https link on sdk

dimasnurenko:
@svor @vparfonov


vparfonov:
ok

PR #113, opened 2015-05-19
IDEX-2560: Added closing opened files after rename module.

AndrienkoAleksandr:


vzhukovskii:
What about ant ptojects?

PR #117, opened 2015-05-21
Removing new popup window for built artifacts.

kaviththiranga:
In chromium based standalone version of Che, opening a new window when downloading the built artifacts causes an additional chromium window being instantiated. To prevent this, we would like to have it opened in the same window,


vparfonov:
fixed here 0a215d54f8192ad68ee7c05108e0d0c212ee073b

PR #131, opened 2015-05-30
Rename the ".codenvy" folder to ".che"

liorokman:
Since Eclipse Che is a part of the Eclipse foundation, the projects folder should be called ".che"
and not ".codenvy".


skabashnyuk:
Unfortunately we are not ready to accept it for 3.x. But we will use it as a base for 4.x. Thanks for contribution.

PR #133, opened 2015-06-01
Align the native git status lists with standard git

tareqhs:
Incorrect behavior in current impl that was fixed:
- doesn't catch conflicting files that are deleted 
- doesn't catch 'added' files if they are later deleted
- adds 'missing' files as 'removed'

Aligning with standard documentation:
http://git-scm.com/docs/git-status#_short_format

Signed-off-by: Tareq Sharafy <tareq.sharafy@sap.com>


benoitf:
@tareqhs There are some tabs characters (it should be space only)

tareqhs:
@benoitf i updated the commit and replaced tabs with spaces.

tareqhs:
d

PR #148, opened 2015-06-18
Demo flux: adding removed char count to document change event

sunix:
Hello,
I need this improvement for real time live editing plugin
This goes with https://github.com/codenvy/che-core/pull/122
Thanks,


vparfonov:
Can you also adopt orion plugin?

sunix:
yes, but at the moment orion plugin change events are not implemented at all. So maybe for another PR

vparfonov:
ok

vparfonov:
In 4.0 version already implement so if we merge your PR to 4.0 we broke the build 
https://github.com/codenvy/che-plugins/blob/4.0/plugin-orion/che-plugin-orion-editor/src/main/java/org/eclipse/che/ide/editor/orion/client/OrionDocument.java#L77

sunix:
OK I will prepare a PR for 4.0

PR #164, opened 2015-07-02
IDEX-2558: added a validation for parameters of remote debug

svor:
@vparfonov


vparfonov:
Ok, need to check filed build on Travis

PR #165, opened 2015-07-05
Bugfix: exceptions in git operations when user profile does not exist

lextiz:
One of flows in function getGitUser did not set users email which is demanded by git plugin, exception was thrown on first git operation if user profile did not exist


skabashnyuk:
Declined due to plugin-git refactoring. See comments in  [che-dev] mail list

PR #168, opened 2015-07-10
IDEX-2350. Added ability to handle each status code when a connection was closed. Show message dialog for Che assembly when a connection was closed.

RomanNikitenko:
Added ability to handle each status code when a connection was closed. 
Show message dialog for Che assembly when a connection was closed.


vparfonov:
ok

PR #172, opened 2015-07-16
Exclude 'equinox' transitive dependency

mkuznyetsov:
Exclude quinox transitive dependency to prevent class loading errors


skabashnyuk:
ok

PR #175, opened 2015-07-20
IDEX-2688: Fixed messages for already existed java source files or pa…

AndrienkoAleksandr:
…ckages. Fixed starting runners on Windows OS.


plugin-java/che-plugin-java-runner-webapps/src/main/java/org/eclipse/che/runner/webapps/WindowsTomcatServer.java
```diff
@@ -179,6 +194,11 @@ public synchronized void stop() throws RunnerException {
             }
             callback.stopped();
             LOG.debug("Stop Tomcat at port {}, application {}", httpPort, workDir);
+            try {
+                logFileConsumer.close();
+            } catch (IOException e) {
+                e.printStackTrace();
+            }
```
skabashnyuk: LOG.error

PR #176, opened 2015-07-21
IDEX-2731. Add formatting source code possibilty.

RomanNikitenko:
Add formatting source code possibilty.


RomanNikitenko:
merged manualy

PR #177, opened 2015-07-21
IDEX-2797: add recipe script into the machine

svor:
@garagatyi @skabashnyuk @evoevodin


garagatyi:
@azatsarynnyy @dimasnurenko work with the client side

PR #187, opened 2015-08-03
IDEX-2666: Fixed run extension on Windows OS

AndrienkoAleksandr:


plugin-sdk/che-plugin-sdk-runner/src/main/java/org/eclipse/che/runner/sdk/CodeServerProcess.java
```diff
@@ -0,0 +1,196 @@
...
+        output.append(String.format("%n====> GWT-code-server.log <===="));
+        output.append(logContent);
+        output.append(System.lineSeparator());
+    }
+
+    private String sendGet(URL url) throws IOException, RunnerException {
+        HttpURLConnection http = null;
+        try {
+            http = (HttpURLConnection)url.openConnection();
+            http.setRequestMethod("GET");
```
skabashnyuk: Read timeout + connection timeout

PR #188, opened 2015-08-03
IDEX-2619:  Added ability to refresh project tree after changing on file system

dimasnurenko:
@evidolob @vparfonov


vparfonov:
merged manually to projectInMachine

PR #193, opened 2015-08-05
IDEX-2717 added docker version check in sdk

akorneta:


plugin-docker/che-plugin-docker-client/src/test/java/org/eclipse/che/plugin/docker/client/DockerVersionTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+import java.io.IOException;
+
+import static org.mockito.Mockito.when;
+
+/**
+ * @author Anton Korneta
+ */
+@Listeners(MockitoTestNGListener.class)
+public class DockerVersionTest {
```
garagatyi: You are testing verifier, so I propose rename test to DockerVersionVerifierTest

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerVersionVerifier.java
```diff
@@ -0,0 +1,56 @@
...
+    public DockerVersionVerifier(DockerConnector dockerConnector, @Named("machine.supported_docker_version") String[] supportedVersions) {
+        this.dockerConnector = dockerConnector;
+        this.supportedVersions = Sets.newHashSet(supportedVersions);
+    }
+
+    /**
+     * Check docker version compatibility.
+     */
+    @PostConstruct
+    public void checkCompatibility() throws ServerException {
```
garagatyi: why is it public? It should be at least package-private

PR #195, opened 2015-08-11
IDEX-2798. Remove unnecessary call hide() of the AsyncRequestLoader.

RomanNikitenko:
Remove unnecessary call hide() of the AsyncRequestLoader.


vparfonov:
ok

RomanNikitenko:
merged manualy

PR #197, opened 2015-08-13
IDEX-2842 Move Docker client and runner from plugins to che-plugins

benoitf:


garagatyi:
For docker API client:
you've created package org.eclipse.che.docker
whereas in 4.0 package is org.eclipse.che.plugin.docker

can you change your package to make master and 4.0 synchronous

The same for docker runner and che-plugin-docker-ext-client.

benoitf:
ok, refactoring done

plugin-docker/che-plugin-docker-client/src/main/resources/org/eclipse/che/plugin/docker/client/dto/Dto.gwt.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
garagatyi: Why you need client side gwt file in docker API client?

plugin-docker/che-plugin-docker-client/pom.xml
```diff
@@ -0,0 +1,180 @@
...
+            <groupId>org.bouncycastle</groupId>
+            <artifactId>bcpkix-jdk15on</artifactId>
+            <version>${org.bouncycastle.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-api-core</artifactId>
+            <version>${che.core.version}</version>
+        </dependency>
+        <dependency>
```
garagatyi: why we need gwt here?

plugin-docker/pom.xml
```diff
@@ -0,0 +1,46 @@
...
+    <artifactId>che-plugin-docker-parent</artifactId>
+    <packaging>pom</packaging>
+    <name>Che Plugin :: Docker :: Parent</name>
+    <modules>
+        <module>che-plugin-docker-client</module>
+        <module>che-plugin-docker-runner</module>
+        <module>che-plugin-docker-ext-client</module>
+    </modules>
+    <properties>
+        <dto-generator-out-directory>${project.build.directory}/generated-sources/dto/</dto-generator-out-directory>
```
garagatyi: why we need dto here?

plugin-docker/pom.xml
```diff
@@ -0,0 +1,46 @@
...
+        <gwt.test.skip>true</gwt.test.skip>
+    </properties>
+    <build>
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>org.eclipse.che.core</groupId>
```
garagatyi: why we need dto maven plugin here?

benoitf:
So, to sum up quickly.
There are DTO because the ext-client is using these DTO
And Client DTO are requiring GWT

skabashnyuk:
ok

PR #201, opened 2015-08-18
IDEX-2905 Allow to use Docker Machine system variables (master branch)

benoitf:
Variables define the boot2docker configuration (like ip address, protocol, certificates folder, etc)

### DIFF
--- a/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
+++ b/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
@@ -14,21 +14,40 @@ import com.google.inject.Inject;
 import com.google.inject.name.Named;
 
 import org.eclipse.che.api.core.util.SystemInfo;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import javax.annotation.Nonnull;
 import java.io.File;
 import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Map;
+
+import static org.eclipse.che.plugin.docker.client.DockerConnector.DEFAULT_DOCKER_MACHINE_CERTS_DIR;
+import static org.eclipse.che.plugin.docker.client.DockerConnector.DEFAULT_DOCKER_MACHINE_URI;
+import static org.eclipse.che.plugin.docker.client.DockerConnector.DOCKER_CERT_PATH_PROPERTY;
+import static org.eclipse.che.plugin.docker.client.DockerConnector.DOCKER_HOST_PROPERTY;
+import static org.eclipse.che.plugin.docker.client.DockerConnector.DOCKER_TLS_VERIFY_PROPERTY;
+import static org.eclipse.che.plugin.docker.client.DockerConnector.UNIX_SOCKET_URI;
 
 /**
  * @author Alexander Garagatyi
+ * @author Florent Benoit
  */
 public class DockerConnectorConfiguration {
+
+    /**
+     * Logger.
+     */
+    private static final Logger LOG = LoggerFactory.getLogger(DockerConnectorConfiguration.class);
+
     @Inject(optional = true)
     @Named("docker.client.daemon_url")
     private URI dockerDaemonUri = dockerDaemonUri();
 
     @Inject(optional = true)
     @Named("docker.client.certificates_folder")
-    private String dockerCertificatesDirectoryPath = boot2dockerCertsDirectoryPath();
+    private String dockerCertificatesDirectoryPath = dockerMachineCertsDirectoryPath();
 
     @Inject
     private InitialAuthConfig authConfigs;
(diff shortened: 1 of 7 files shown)


plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
```diff
@@ -47,11 +63,57 @@ public DockerConnectorConfiguration(URI dockerDaemonUri,
...
+                // unable to use given property, fallback to default URL
+                return DEFAULT_BOOT2DOCKER_URI;
+
+            }
+        }
+
+
+
+
+
```
evoevodin: Generally it is okay, but come on is it really necessary to have 5 empty lines here?

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
```diff
@@ -14,14 +14,30 @@
...
+import static org.eclipse.che.plugin.docker.client.DockerConnector.DEFAULT_BOOT2DOCKER_URI;
+import static org.eclipse.che.plugin.docker.client.DockerConnector.UNIX_SOCKET_URI;
 
 /**
  * @author Alexander Garagatyi
  */
 public class DockerConnectorConfiguration {
+
+    /**
+     * Logger.
```
evoevodin: We are always welcome for javadoc, but what the sense of this one?

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
```diff
@@ -14,14 +14,30 @@
...
+
+import static org.eclipse.che.plugin.docker.client.DockerConnector.BOOT2DOCKER_CERT_PATH_PROPERTY;
+import static org.eclipse.che.plugin.docker.client.DockerConnector.BOOT2DOCKER_HOST_PROPERTY;
+import static org.eclipse.che.plugin.docker.client.DockerConnector.BOOT2DOCKER_TLS_VERIFY_PROPERTY;
+import static org.eclipse.che.plugin.docker.client.DockerConnector.DEFAULT_BOOT2DOCKER_CERTS_DIR;
+import static org.eclipse.che.plugin.docker.client.DockerConnector.DEFAULT_BOOT2DOCKER_URI;
+import static org.eclipse.che.plugin.docker.client.DockerConnector.UNIX_SOCKET_URI;
 
 /**
  * @author Alexander Garagatyi
```
evoevodin: Can you please add yourself as a <code>@author</code>

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
benoitf: well I comment all what I add in general. Looks more pretty. Checkstyle or sonar should be happier as well. I should remove it ?

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
benoitf: oops. Will fix it asap

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
benoitf: ok

skabashnyuk: +1

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
skabashnyuk: yes

benoitf: yes I should remove ? not a joke ?

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
benoitf: done

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
benoitf: done

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
skabashnyuk: Up to you. You can leave it as is. But it's a dummy comment - no?

benoitf: OK, I prefer to leave it as this. Just habit to have everything commented on top of method and inside methods describing logic

skabashnyuk:
ok

benoitf:
Rethinking about it, I may add some tests but I would require to mock System.getEnv() (^^)
So instead of trying to mock java.lang, I would need to add a new method that carry on the Environment. Would it be OK then ?

something like   private static URI dockerDaemonUri() 
    redirecting to dockerDaemonUri(System.getEnv()))

and another method dockerDaemonUri that will take Map<String, String>

benoitf:
Ok I've added tests

evoevodin:
It seems everything is okay, well done :+1:

benoitf:
Also I add PR for 4.0 branch
https://github.com/codenvy/che-plugins/pull/206

skabashnyuk:
ok go

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
garagatyi: I also dislike useless and obvious comments. And I remove such comments when I see them. Please, do not commit such things anymore. By the way see logger declaration standart https://wiki.codenvycorp.com/display/DPP/Logger

benoitf: If it's final and static, it should be in uppercase as well
LOGGER and not logger to be compliant.

About comments, you may dislike it but I'm conforming to code standard http://www.oracle.com/technetwork/articles/java/index-137868.html.

benoitf: Also I don't like when interface are not commented as well
looking at https://github.com/codenvy/che-plugins/blob/4.0/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineFactory.java

PR #208, opened 2015-08-24
IDEX-2946 Use the URL by relying on injection without hardcoding path

benoitf:


skabashnyuk:
ok

benoitf:
backported in 4.0 as well

PR #214, opened 2015-08-27
IDEX-2964 Allow to use server-side only extension 

benoitf:
(in addition to client only and client+server extensions)


vparfonov:
ok for me

PR #217, opened 2015-08-28
IDEX-2180:added an ability to save git committer name and email into user prefe rences

svor:


svor:
Merged into 4.0. [Git commit](https://github.com/codenvy/che-plugins/commit/5a867109276e8d890ff86e2952ef6eb12c6ff25b)

PR #219, opened 2015-08-31
IDEX-2899. Added RemoveMavenModuleHandler for remove module from pom.xml. Fixed bug in Model at removing module.

RomanNikitenko:
Added RemoveMavenModuleHandler for remove module from pom.xml. 
Fixed bug in Model at removing module.


plugin-java/che-plugin-java-maven-tools/src/main/java/org/eclipse/che/ide/maven/tools/Model.java
```diff
@@ -1061,7 +1061,10 @@ private void removeModuleFromXML(String module) {
         if (modules.isEmpty()) {
             root.removeChild("modules");
         } else {
-            for (Element element : root.getChildren()) {
+            if (!root.hasSingleChild("modules")) {
```
evoevodin: Why do we need this check?

The only place where this method is used is [here](https://github.com/codenvy/che-plugins/blob/IDEX-2899/plugin-java/che-plugin-java-maven-tools/src/main/java/org/eclipse/che/ide/maven/tools/Model.java#L548), as you can see this method will be invoked if and only if <code>modules().remove(module)</code> returns true. 
Also there is no way to add module without affecting of XML content, which is also important and actually means that when <code>modules()</code> returns non empty list XML content contains non empty modules as well

evoevodin:
Can you please add tests

RomanNikitenko:
merged manually

PR #220, opened 2015-09-01
IDEX-2855: Remove extensions servers info from non dev machines

garagatyi:
@skabashnyuk @evoevodin


plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -259,18 +275,39 @@ private Instance createInstance(String imageId,
...
                                     Recipe recipe,
                                     int memorySizeMB,
                                     LineConsumer outputConsumer)
             throws MachineException {
         try {
+            final Map<String, String> labels;
+            final Map<String, Map<String, String>> portsToExpose;
+            final Set<String> volumes;
+            if (isDev) {
+                labels = new HashMap<>(machineContainerLabels.size() + devMachineContainerLabels.size());
```
evoevodin: What do you think about <code>Maps.newHashMapWithExpectedSize(...)</code> in this case?

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -259,18 +275,39 @@ private Instance createInstance(String imageId,
...
+                labels.putAll(machineContainerLabels);
+                labels.putAll(devMachineContainerLabels);
+
+                portsToExpose = new HashMap<>(portsToExposeOnMachine.size() + portsToExposeOnDevMachine.size());
+                portsToExpose.putAll(portsToExposeOnMachine);
+                portsToExpose.putAll(portsToExposeOnDevMachine);
+
+                // 1 extra element that contains workspace FS folder will be added further
+                volumes = new HashSet<>(devMachineSystemVolumes.size() + systemVolumesForMachine.size() + 1);
```
evoevodin: What about <code>Sets.newHashSetWithExpectedSize(...)</code> in this case?

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
garagatyi: +1 10x

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
garagatyi: +1

skabashnyuk:
LGTM

PR #224, opened 2015-09-02
IDEX-2970 Docker recipes should be packaged along the docker plugin

benoitf:
Move it from che/recipes


benoitf:
@skabashnyuk @garagatyi 
It's just a move

skabashnyuk:
ok for me

PR #228, opened 2015-09-04
Idex 2790. Initialize extServerStateController when dev machine started.

RomanNikitenko:
Idex 2790. Initialize extServerStateController when dev machine started.


vparfonov:
ок

PR #230, opened 2015-09-04
IDEX-3003: add endpoint url as env variable in dev machine

garagatyi:
@skabashnyuk @evoevodin 
Please, review.

### DIFF
--- a/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ext/LocalStorageDockerVolumePathProvider.java
+++ b/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ext/LocalStorageDockerVolumePathProvider.java
@@ -1,0 +1,34 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2015 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.plugin.docker.machine.ext;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Provider;
+
+/**
+ * Provides volumes configuration of machine for terminal
+ *
+ * @author Alexander Garagatyi
+ */
+public class LocalStorageDockerVolumePathProvider implements Provider<String> {
+    private final String localStoragePath;
+
+    @Inject
+    public LocalStorageDockerVolumePathProvider(@Named("local.storage.path") String localStoragePath) {
+        this.localStoragePath = localStoragePath;
+    }
+
+    @Override
+    public String get() {
+        return localStoragePath + ":/local-storage";
+    }
+}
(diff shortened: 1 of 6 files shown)


skabashnyuk:
:clap:

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ext/LocalStorageDockerVolumePathProvider.java
```diff
@@ -0,0 +1,34 @@
...
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.plugin.docker.machine.ext;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Provider;
+
+/**
+ * Provides volumes configuration of machine for terminal
```
benoitf: Strange documentation
For me it's the same than this one
https://github.com/codenvy/che-plugins/blob/4.0/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ext/TerminalServerBindingProvider.java#L18

garagatyi: it's copy-paste error.

garagatyi: >javadoc is in option ?
@benoitf What do you mean?

benoitf: My meaning is that the code should be more documented for someone that want to read it.

public ContainerCreated(String id, String[] warnings) { <-- what are these warnings

public static final String API_ENDPOINT_URL_VARIABLE = "CHE_API_ENDPOINT";
what is the purpose of this constant : why the name is different of the content ?

LocalStorageDockerVolumePathProvider : wrong documentation

Also on a separated note, why servicetest is modified ? 
what is the relation of the issue with that change ? it seems to be a cleanup and a change from tmp to fake. Looking at the history we may wonder what the relation of this change and the name of the commit

also as there was a commit related to this PR, why status is not merged  and closed ?

garagatyi: >public ContainerCreated(String id, String[] warnings) { <-- what are these warnings

I just added not default constructor to existing class. This class was added by Andrew a long time ago. I didn't find description of this field in docker API.

>public static final String API_ENDPOINT_URL_VARIABLE = "CHE_API_ENDPOINT";
what is the purpose of this constant : why the name is different of the content ?

Please take a look at IDEX-3003. I suppose it helps.

>LocalStorageDockerVolumePathProvider : wrong documentation

I've fixed it.

>Also on a separated note, why servicetest is modified ? 

Because constructor of the class that is used in this test was changed.

>what is the relation of the issue with that change ? it seems to be a cleanup and a change from tmp to fake. Looking at the history we may wonder what the relation of this change and the name of the commit

Class didn't have tests at all, so I created tests. I agree that it would be better to commit other test methods with separate commit, my fault.

>also as there was a commit related to this PR, why status is not merged and closed ?

I merged it manually to pretify git history.

PR #232, opened 2015-09-09
IDEX-2992. Changed machine status channel. Fixed refresh list of machines and menu 'Machine' after destroy last machine.

RomanNikitenko:
Changed machine status channel. 
Fixed refresh list of machines and menu 'Machine' after destroy last machine.


vparfonov:
ok

PR #236, opened 2015-09-10
IDEX-2971: Tree collapse fixes

vzhukovskii:
@vparfonov


plugin-github/che-plugin-github-ext-github/src/main/java/org/eclipse/che/ide/ext/github/client/importer/page/GithubImporterPagePresenter.java
```diff
@@ -169,9 +171,13 @@ public void keepDirectorySelected(boolean keepDirectory) {
     public void keepDirectoryNameChanged(@Nonnull String directoryName) {
         if (view.keepDirectory()) {
             projectParameters().put("keepDirectory", directoryName);
+            dataObject.getProject().setContentRoot(view.getDirectoryName());
```
skabashnyuk: may be we should introduce local variable here ?

PR #239, opened 2015-09-11
IDEX-3038; chande namings from codenvy to che;

mshaposhnik:


sleshchenko:
What is about https://github.com/codenvy/che-plugins/pull/239/files#diff-f0ecf8ed5a42b06585a50e5e462750d5L87?

sleshchenko:
I think you also should change https://github.com/codenvy/che-core/blob/master/che-core-git-server/src/main/java/com/codenvy/ide/git/VFSPermissionsFilter.java#L116 and  ttps://github.com/codenvy/che-core/blob/master/che-core-git-impl-native/src/main/java/org/eclipse/che/git/impl/nativegit/NativeGitConnection.java#L691

PR #242, opened 2015-09-12
Idex 2376

garagatyi:
@skabashnyuk @evoevodin @akorneta


plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -566,90 +517,84 @@ public String commit(String container, String repository, String tag, String com
...
-            final ExecStart execStart = new ExecStart().withDetach(execOutputProcessor == null);
-            final String entity = JsonHelper.toJson(execStart, FIRST_LETTER_LOWERCASE);
-            headers.add(Pair.of("Content-Length", entity.getBytes().length));
-            final DockerResponse response = connection.method("POST")
-                                                      .path(String.format("/exec/%s/start", execId))
-                                                      .headers(headers)
-                                                      .entity(entity)
-                                                      .request();
+    public void startExec(String execId, MessageProcessor<LogMessage> execOutputProcessor) throws IOException {
+        final List<Pair<String, ?>> headers = new ArrayList<>(2);
```
evoevodin: I think in this case it is better to define <code>headers</code> variable above <code>580</code> line + move 
<code>headers.add(Pair.of("Content-Type", "application/json"));</code> after that declaration, then it will be like this:
```java
final List<Pair<String, ?>> headers = new ArrayList<>(2);
headers.add(Pair.of("Content-Type", "application/json"));
headers.add(Pair.of("Content-Length", entity.getBytes().length));
```

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/CgroupOOMDetector.java
```diff
@@ -0,0 +1,269 @@
...
+
+    @Inject
+    public CgroupOOMDetector(DockerConnectorConfiguration connectorConfiguration, DockerConnector dockerConnector) {
+        this(connectorConfiguration.getDockerDaemonUri(), dockerConnector);
+    }
+
+    public CgroupOOMDetector(URI dockerDaemonUri, DockerConnector dockerConnector) {
+        this.dockerDaemonUri = dockerDaemonUri;
+        this.dockerConnector = dockerConnector;
+        oomDetectors = new ConcurrentHashMap<>();
```
skabashnyuk: this

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/CgroupOOMDetector.java
```diff
@@ -0,0 +1,269 @@
...
+        executor = Executors.newCachedThreadPool(new ThreadFactoryBuilder().setNameFormat("CgroupOOMDetector-%d").setDaemon(true).build());
```
skabashnyuk: this

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/CgroupOOMDetector.java
```diff
@@ -0,0 +1,269 @@
...
+    public void startDetection(String container, MessageProcessor<LogMessage> containerLogProcessor) {
+        if (needStartOOMDetector()) {
+            if (cgroupMount == null) {
+                LOG.warn("System doesn't support OOM events");
+                return;
+            }
+            try {
+                final long memory = dockerConnector.inspectContainer(container).getConfig().getMemory();
+                final OOMDetector oomDetector = new OOMDetector(container, containerLogProcessor, memory);
+                oomDetectors.put(container, oomDetector);
```
skabashnyuk: not sure it's thread safe

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/CgroupOOMDetector.java
```diff
@@ -0,0 +1,269 @@
...
+     * So it's complicated to detect time when oom-killer was activated and link its activity with failed docker container.
+     *
+     *     2. Usage of cgroup notification mechanism.
+     * Good article about this: https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/6/html/Resource_Management_Guide/sec-Using_the_Notification_API.html
+     */
+    private static String  cgroupMount;
+    private static boolean systemd;
+
+    static {
+        if (SystemInfo.isLinux()) {
```
skabashnyuk: what about osx?

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/CgroupOOMDetector.java
```diff
@@ -0,0 +1,269 @@
...
+            final String mounts = "/proc/mounts";
+            try (BufferedReader reader = Files.newBufferedReader(Paths.get(mounts), Charset.forName("UTF-8"))) {
```
skabashnyuk: may be stream API?

```java
        try (BufferedReader reader = new BufferedReader(source)) {
            return reader.lines()
                    .substream(1)
                    .map(line -> Arrays.asList(line.split(separator)))
                    .collect(Collectors.toList());
        } catch (IOException e) {
            throw new UncheckedIOException(e);
        }
```

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -660,53 +605,46 @@ public void startExec(String execId, LogMessageProcessor execOutputProcessor) th
...
-                    pathBuilder.append(URLEncoder.encode(psArgs[i], "UTF-8"));
+        final List<Pair<String, ?>> headers = new ArrayList<>(2);
+        headers.add(Pair.of("Content-Type", "text/plain"));
+        headers.add(Pair.of("Content-Length", 0));
+        final DockerConnection connection = openConnection(dockerDaemonUri).method("GET")
+                                                                           .path("/containers/" + container + "/top")
+                                                                           .headers(headers);
+        if (psArgs != null && psArgs.length != 0) {
+            StringBuilder psArgsQueryBuilder = new StringBuilder();
+            for (int i = 0, l = psArgs.length; i < l; i++) {
```
evoevodin: As far as i understand this loop maybe replaced with <code>Arrays.stream</code> + <code>Stream.map</code> + <code>Collectors.joining</code>.

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -366,22 +333,18 @@ public void startContainer(String container, HostConfig hostConfig, LogMessagePr
...
-        final DockerConnection connection = openConnection(dockerDaemonUri);
-        try {
-            final List<Pair<String, ?>> headers = new ArrayList<>(2);
-            headers.add(Pair.of("Content-Type", "text/plain"));
-            headers.add(Pair.of("Content-Length", 0));
-            final DockerResponse response =
-                    connection.method("POST").path(String.format("/containers/%s/stop?t=%d", container, timeunit.toSeconds(timeout)))
-                              .headers(headers).request();
+        final List<Pair<String, ?>> headers = new ArrayList<>(2);
+        headers.add(Pair.of("Content-Type", "text/plain"));
```
skabashnyuk: may constant "text/plain"?

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -395,22 +358,19 @@ public void stopContainer(String container, long timeout, TimeUnit timeunit) thr
...
-        final DockerConnection connection = openConnection(dockerDaemonUri);
-        try {
-            final List<Pair<String, ?>> headers = new ArrayList<>(2);
-            headers.add(Pair.of("Content-Type", "text/plain"));
-            headers.add(Pair.of("Content-Length", 0));
-            final DockerResponse response = connection.method("POST")
-                                                      .path(String.format("/containers/%s/kill?signal=%d", container, signal))
-                                                      .headers(headers).request();
+        final List<Pair<String, ?>> headers = new ArrayList<>(2);
+        headers.add(Pair.of("Content-Type", "text/plain"));
```
skabashnyuk: constant?

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -525,27 +474,29 @@ protected ContainerInfo doInspectContainer(String container, URI dockerDaemonUri
...
-        try {
-            final List<Pair<String, ?>> headers = new ArrayList<>(2);
-            headers.add(Pair.of("Content-Type", "text/plain"));
-            headers.add(Pair.of("Content-Length", 0));
-            final String path = String.format("/containers/%s/attach?stream=%d&logs=%d&stdout=%d&stderr=%d", container, (stream ? 1 : 0),
-                                              (stream ? 0 : 1), 1, 1);
-            final DockerResponse response = connection.method("POST").path(path).headers(headers).request();
+    public void attachContainer(String container, MessageProcessor<LogMessage> containerLogsProcessor, boolean stream) throws IOException {
+        final List<Pair<String, ?>> headers = new ArrayList<>(2);
+        headers.add(Pair.of("Content-Type", "text/plain"));
```
skabashnyuk: constant?

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -566,90 +517,84 @@ public String commit(String container, String repository, String tag, String com
...
-            final ExecConfig execConfig = new ExecConfig().withCmd(cmd);
-            if (!detach) {
-                execConfig.withAttachStderr(true).withAttachStdout(true);
-            }
-            final String entity = JsonHelper.toJson(execConfig, FIRST_LETTER_LOWERCASE);
-            headers.add(Pair.of("Content-Length", entity.getBytes().length));
-            final DockerResponse response = connection.method("POST").path(String.format("/containers/%s/exec", container))
-                                                      .headers(headers).entity(entity).request();
+        final List<Pair<String, ?>> headers = new ArrayList<>(2);
+        headers.add(Pair.of("Content-Type", "application/json"));
```
skabashnyuk: constant?

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -660,53 +605,46 @@ public void startExec(String execId, LogMessageProcessor execOutputProcessor) th
...
-            } else {
-                final StringBuilder pathBuilder = new StringBuilder();
-                pathBuilder.append("/containers/").append(container).append("/top?ps_args=");
-                for (int i = 0, l = psArgs.length; i < l; i++) {
-                    if (i > 0) {
-                        pathBuilder.append('+');
-                    }
-                    pathBuilder.append(URLEncoder.encode(psArgs[i], "UTF-8"));
+        final List<Pair<String, ?>> headers = new ArrayList<>(2);
+        headers.add(Pair.of("Content-Type", "text/plain"));
```
skabashnyuk: constant?

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -765,28 +706,72 @@ public void putResource(String container, String targetPath, InputStream sourceS
...
         } finally {
-            if (connection != null) {
-                connection.close();
-            }
             FileCleaner.addFile(tarFile);
         }
     }
 
     /**
+     * Get docker events.
```
evoevodin: I think it is not formatted, blank line is required after this description.

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -813,165 +798,161 @@ protected String doBuildImage(String repository,
...
-            final List<Pair<String, ?>> headers = new ArrayList<>(3);
-            headers.add(Pair.of("Content-Type", "application/x-compressed-tar"));
-            headers.add(Pair.of("Content-Length", tar.length()));
-            headers.add(Pair.of("X-Registry-Config", Base64.encodeBase64String(JsonHelper.toJson(authConfigs).getBytes())));
-            final DockerResponse response;
-            try (InputStream tarInput = new FileInputStream(tar)) {
-                response = connection.method("POST").path(String.format("/build?t=%s&rm=%d&pull=%d", repository, 1, 1)).headers(headers)
-                                     .entity(tarInput).request();
+        final List<Pair<String, ?>> headers = new ArrayList<>(3);
+        headers.add(Pair.of("Content-Type", "application/x-compressed-tar"));
```
skabashnyuk: ExtMadiaType?

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/JsonMessageReader.java
```diff
@@ -0,0 +1,70 @@
...
+ * @author Alexander Garagatyi
+ */
+public class JsonMessageReader<T> {
+    private static final Gson GSON = new Gson();
+
+    private final JsonStreamParser    streamParser;
+    private final Class<T>            messageClass;
+    private final PushbackInputStream inputStream;
+
+    private boolean firstRead = false;
```
evoevodin: <code>firstRead</code> is always <code>false</code>

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/JsonMessageReader.java
```diff
@@ -0,0 +1,70 @@
...
+     *                     we can't get parameter class of current class.
+     */
+    public JsonMessageReader(InputStream source, Class<T> messageClass) {
+        this.inputStream = new PushbackInputStream(source);
+        this.streamParser = new JsonStreamParser(new InputStreamReader(source));
+        this.messageClass = messageClass;
+    }
+
+    public T next() throws IOException {
+        if (firstRead) {
```
evoevodin: It is always <code>false</code> see [comment](https://github.com/codenvy/che-plugins/pull/242/files#r39343330)

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/JsonMessageReader.java
```diff
@@ -0,0 +1,70 @@
...
+            int firstByte = inputStream.read();
+            if (firstByte == -1) {
+                return null;
+            } else {
```
evoevodin: Don't you think that in this case is better to avoid else statement?
```java
int firstByte = inputStream.read();
if (firstByte == -1) {
    return null;
}
inputStream.unread(firstByte);
firstRead = false;
```

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/JsonMessageReader.java
```diff
@@ -0,0 +1,70 @@
...
+import com.google.gson.JsonStreamParser;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.PushbackInputStream;
+
+/**
+ * Docker daemon sends chunked data in response. One chunk isn't always one JSON object so need to read full chunk at once to be able
+ * restore JSON object. This reader merges (if needs) few chunks until get full JSON object that we can parse.
```
evoevodin: I think this documentation is not correct at all, as i can see this class reads _json_ objects _one by one_ from input stream(if possible to do so) and tries to parse it into <code>T</code>, i mean it doesn't merge anything. 
 Also what is _full JSON object_?

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/JsonMessageReader.java
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * @author Alexander Garagatyi
+ */
+public class JsonMessageReader<T> {
```
evoevodin: You must document <code>T</code> as well.

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/JsonMessageReader.java
```diff
@@ -0,0 +1,70 @@
...
+    private static final Gson GSON = new Gson();
+
+    private final JsonStreamParser    streamParser;
+    private final Class<T>            messageClass;
+    private final PushbackInputStream inputStream;
```
evoevodin: The usage of <code>PushbackInputStream</code> is not pretty obvious, i think it should be documented in this place or where it used.

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/JsonMessageReader.java
```diff
@@ -0,0 +1,70 @@
...
+     *                     Because of erasure of generic information in runtime in some cases
+     *                     we can't get parameter class of current class.
+     */
+    public JsonMessageReader(InputStream source, Class<T> messageClass) {
+        this.inputStream = new PushbackInputStream(source);
+        this.streamParser = new JsonStreamParser(new InputStreamReader(source));
+        this.messageClass = messageClass;
+    }
+
+    public T next() throws IOException {
```
evoevodin: It should be documented, at least cases when it returns <code>null</code> and when instance of <code>T</code>

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/connection/DockerConnection.java
```diff
@@ -39,6 +42,25 @@ public DockerConnection path(String path) {
         return this;
     }
 
+    public DockerConnection query(String name, Object... values) {
+        if (name == null) {
+            throw new IllegalArgumentException("Name is null");
```
evoevodin: <code>NullPointerException</code> must be thrown in this case, see _60 Item_ from _Effective Java 2nd ed_.
Basically it says:
>If a caller passes null in some parameter for which null values are prohibited, convention dictates that NullPointerException be thrown rather than IllegalArgumentException.

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/connection/DockerConnection.java
```diff
@@ -60,10 +82,11 @@ public DockerConnection entity(byte[] entity) {
     }
 
     public DockerResponse request() throws IOException {
-        return request(method, path, headers, entity);
+        return request(method, path, query.toString(), headers, entity);
     }
 
-    protected abstract DockerResponse request(String method, String path, List<Pair<String, ?>> headers, Entity entity) throws IOException;
+    protected abstract DockerResponse request(String method, String path, String query, List<Pair<String, ?>> headers, Entity entity)
```
evoevodin: As for me
```java
 protected abstract DockerResponse request(String method, 
                                           String path, 
                                           String query, 
                                           List<Pair<String, ?>> headers, 
                                           Entity entity) throws IOException {
```
looks better

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/Filters.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Representation of docker filters.
+ *
+ * @author Alexander Garagatyi
+ */
+public class Filters {
+    private final Map<String, List<String>> filters = new HashMap<>();
```
evoevodin: What do you think about <code>ListMultimap</code> + return <code> ImmutableListMultimap</code> from <code>getFilters</code> method ?
Then you can avoid map copies.

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/CgroupOOMDetector.java
garagatyi: no support for OSX

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
garagatyi: Do you propose to use constant with the same value?

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
garagatyi: agree

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/JsonMessageReader.java
garagatyi: Full JSON here means valid JSON. Chunk may end with not valid JSON. To make data valid JSON it is needed to read next chunk, concatenate it with the previous one and then try to parse into JSON. 
By the way I did not change the logic of JSON parsing, just moved code to this class and made parsing parameterized.
Authors of this code are Andrew and you. And comment was written by someone of you.

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/JsonMessageReader.java
garagatyi: agree

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/JsonMessageReader.java
garagatyi: agree

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/JsonMessageReader.java
garagatyi: yep, it's a bug. This line should initialize variable with value `true`

garagatyi: thank you

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/JsonMessageReader.java
garagatyi: agree

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/connection/DockerConnection.java
garagatyi: agree

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/connection/DockerConnection.java
garagatyi: agree

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/Filters.java
garagatyi: +1 I will try

plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/ServiceTest.java
```diff
@@ -129,6 +136,14 @@ public void tearDownClass() throws IOException {
         docker.removeContainer(registryContainerId, true, false);
     }
 
+//    @Test
```
garagatyi: oops was added by mistake

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/CgroupOOMDetector.java
garagatyi: I would not want to change everything when I just move code from one class to another. It can be done later.

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
garagatyi: URLEncoder.encode throws UnsupportedEncodingException, so lambda won't be pretty. WDYT?

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
garagatyi: Next several lines are part of the description

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
evoevodin: Hmm, i see. 
What do you think about extracting <code>URLEncoder#encode</code> into method which rethrows <code>IOException</code> as </code>RuntimeException</code> and use method reference of it, OR as far as i remember <code>UrlEscapers</code> from _guava_ may help you in this case.

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
garagatyi: Value in this line differs from the value in ExtMediaType

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
evoevodin: Then its okay. As for me it is just better to separate _Summary_ from _Description_ with 1 blank line.

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/Filters.java
garagatyi: We use reflection to convert model objects to JSONs and vise versa. So it is possible that another method of docker API will use this object as part of another object. That's why I would avoid `ListMultimap` here

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/CgroupOOMDetector.java
skabashnyuk: is this works inside of the machine

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/CgroupOOMDetector.java
skabashnyuk: ok

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
skabashnyuk: I propose to use constant

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/CgroupOOMDetector.java
garagatyi: nope

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
garagatyi: I think it can be done later, especially if you would like to do that :)

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
garagatyi: done

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
garagatyi: done

skabashnyuk:
LGTM

PR #246, opened 2015-09-14
IDEX-2964 Allow to add server side extension only (without client side)

benoitf:


benoitf:
It's like for https://github.com/codenvy/che-plugins/commit/370e6b12eb37964d7d0e42605c6cdbb467c79ed3 as code seems to be duplicated

PR #251, opened 2015-09-17
IDEX-3066 use correct bind volume dir on Windows

benoitf:


skabashnyuk:
ok

plugin-docker/che-plugin-docker-runner/src/main/java/org/eclipse/che/plugin/docker/runner/BaseDockerRunner.java
```diff
@@ -319,7 +319,18 @@ protected ApplicationProcess newApplicationProcess(DeploymentSources toDeploy, R
                         unpackedApplication = unpackArchive(application, workDir, applicationFilename + "_unpack");
                     }
                 }
-                hostConfig.setBinds(new String[]{String.format("%s:%s", unpackedApplication.getAbsolutePath(), applicationBindDir)});
+
+                // On Windows binding directory needs to follow URL convention with first / and no colon :
+                // instead of C:\\Users\\user it needs to be /c/Users/user (note as well the lowercase c at first)
```
garagatyi: it would be better to avoid slash symbol escaping in the comments, because it is not obvious why two slashes are used here

plugin-docker/che-plugin-docker-runner/src/main/java/org/eclipse/che/plugin/docker/runner/BaseDockerRunner.java
```diff
@@ -319,7 +319,18 @@ protected ApplicationProcess newApplicationProcess(DeploymentSources toDeploy, R
                         unpackedApplication = unpackArchive(application, workDir, applicationFilename + "_unpack");
                     }
                 }
-                hostConfig.setBinds(new String[]{String.format("%s:%s", unpackedApplication.getAbsolutePath(), applicationBindDir)});
+
+                // On Windows binding directory needs to follow URL convention with first / and no colon :
```
garagatyi: Please add link to document that describe that we need to convert path for windows

plugin-docker/che-plugin-docker-runner/src/main/java/org/eclipse/che/plugin/docker/runner/BaseDockerRunner.java
benoitf: I agree but I was using that I got unescaped character as \u is a unicode shortcut

plugin-docker/che-plugin-docker-runner/src/main/java/org/eclipse/che/plugin/docker/runner/BaseDockerRunner.java
benoitf: is https://github.com/boot2docker/boot2docker/blob/master/README.md#virtualbox-guest-additions  ok ?
Also I was using the try and debug mode

plugin-docker/che-plugin-docker-runner/src/main/java/org/eclipse/che/plugin/docker/runner/BaseDockerRunner.java
benoitf: https://docs.oracle.com/javase/specs/jls/se7/html/jls-3.html

plugin-docker/che-plugin-docker-runner/src/main/java/org/eclipse/che/plugin/docker/runner/BaseDockerRunner.java
garagatyi: As I can see it doesn't say that User should be in lowercase. Am I missing something?

plugin-docker/che-plugin-docker-runner/src/main/java/org/eclipse/che/plugin/docker/runner/BaseDockerRunner.java
garagatyi: It doesn't say that comments should use double slashes instead of single, but double slashes make comment less readable

benoitf: Write in any of your java class c:\users, you'll see that you'll a compilation issue as \u will be interpreted as unicode character. And then you'll understand the issue of why in the comment I added \\

plugin-docker/che-plugin-docker-runner/src/main/java/org/eclipse/che/plugin/docker/runner/BaseDockerRunner.java
benoitf: But user is not in lowercase, only c as written in the comment

vparfonov:
ok

plugin-docker/che-plugin-docker-runner/src/main/java/org/eclipse/che/plugin/docker/runner/BaseDockerRunner.java
garagatyi: Oh, sorry, my mistake.

plugin-docker/che-plugin-docker-runner/src/main/java/org/eclipse/che/plugin/docker/runner/BaseDockerRunner.java
garagatyi: I know that in code I have to add backslash, but in comments I don't have to.

benoitf: yes that's odd :-1:

garagatyi:
LGTM

PR #254, opened 2015-09-17
IDEX-2868: Change styles for the command bar

olexii4:


skabashnyuk:
ok for me. What about @vparfonov

PR #256, opened 2015-09-18
IDEX-3099 Memory is not a string but long in hostconfig

benoitf:
From https://docs.docker.com/reference/api/docker_remote_api_v1.20/
to
https://docs.docker.com/reference/api/docker_remote_api_v1.14/
 at least


benoitf:
Else we have this issue
org.eclipse.che.plugin.docker.client.DockerException: Error response from docker API, status: 500, message: json: cannot unmarshal string into Go value of type int64

garagatyi:
LGTM but please divide it into 2 commits, to be able to cherry-pick docker client fix in 4.0

benoitf:
cherry-picked on 4.0 https://github.com/codenvy/che-plugins/commit/bcac2de091234fe83ee47eca095dbcad77241109

PR #258, opened 2015-09-21
IDEX-3104: set cpuShares to "2" as it is minimum allowed starting from docker 1.…

riuvshin:
…7 and above

More information is here: https://github.com/docker/docker/pull/13722


garagatyi:
LGTM

PR #259, opened 2015-09-21
Fix <title> text to say Eclipse Che

TylerJewell:
This is the text that appears in the browser tab when che is loading.


skabashnyuk:
Don't forget to remove your branch

PR #261, opened 2015-09-22
IDEX-3057: Added 'Project configuration' dialog when project has prob…

AndrienkoAleksandr:
IDEX-3057: Added 'Project configuration' dialog when project has problems after git operation 'checkout' and 'reset --hard'


skabashnyuk:
+0

PR #264, opened 2015-09-23
IDEX-1979:added refresh remote branches field in 'Fetch' and 'Pull' forms

vinokurig:


skabashnyuk:
+0
@vparfonov ?

plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/fetch/FetchPresenter.java
```diff
@@ -78,20 +78,20 @@ public void showDialog() {
...
+        setRemotes();
     }
 
     /**
-     * Get the list of remote repositories for local one. If remote repositories are found, then get the list of branches (remote and
+     * Set the list of remote repositories for local one. If remote repositories are found, then set the list of branches (remote and
      * local).
      */
-    private void getRemotes() {
+    private void setRemotes() {
```
sleshchenko: Why do you choose setRemotes name for this method. In general method with name set should have some parameters which it will be set. I think this method should have name something like get or fetch or update, but not set. @vinokurig WDYT?

vparfonov:
ok

vinokurig:
merged manually with --squash to 7b6797952d3edd22ede210e9ba83d9e7892fb6d3

PR #272, opened 2015-09-25
IDEX-2990 tree bugfixes and improvements

vzhukovskii:
@vparfonov @skabashnyuk 

- added "scroll from source" button, like in IDEA
- fixed refreshing project explorer after creating module
- refactor delete item action to allow work through promises
- added ability to handle delete key
- refactored delete dialog to show additional information about deleted items like in IDEA
- added additional validation to item rename, so that item cannot be renamed into self name
- improved project explorer refreshing files mechanism after renaming folder with similar name
- added additional event SynchronizeProjectViewEvent to allow reload files throgh eventbus
- fix selection changing after right click on the node
- move NewProjectExplorer* names into persisted ProjectExplorer* names


skabashnyuk:
ok for me

vparfonov:
ok

PR #274, opened 2015-09-25
IDEX-916: added progress status to some operations

vinokurig:


plugin-java/che-plugin-java-ext-java/src/main/java/org/eclipse/che/ide/ext/java/client/dependenciesupdater/DependenciesUpdater.java
```diff
@@ -110,7 +110,7 @@ public void updateDependencies(final ProjectDescriptor project, final boolean fo
 
         builderConsole.clear();
 
-        notification = new Notification(javaLocalizationConstant.updatingDependencies(), PROGRESS);
+        notification = new Notification(javaLocalizationConstant.updatingDependencies(), PROGRESS, true);
```
sleshchenko: Can you comment this change, please?

vinokurig: Now "update dependencies" notification is closing by time out, and if the process of updating dependencies is long, the progress notification is closing by time out, but the process is still active. It can be confusing for user that tries to open an editor and it shows error markers because update dependencies is not finished. This change is needed for "update dependencies" notification to be visible all time while the process of updating dependencies is active.

sleshchenko: Nice. Thanks!

PR #278, opened 2015-09-29
Idex 3144

vparfonov:


vparfonov:
WDYT? @skabashnyuk @evoevodin @garagatyi

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -115,6 +117,41 @@ public DockerInstanceProvider(DockerConnector docker,
...
+        return paths.stream().map(s -> escapePath(s)).collect(Collectors.toSet());
+    }
+
+    /**
+     * Escape path for Windows system with boot@docker according to rules given here :
+     * https://github.com/boot2docker/boot2docker/blob/master/README.md#virtualbox-guest-additions
+     * @param path
+     * @return escaped set of path
+     */
+    private String escapePath(String path) {
```
evoevodin: Looks like an util method which we may need in the future.
Maybe we should move it to _Util_ class and also test.

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -115,6 +117,41 @@ public DockerInstanceProvider(DockerConnector docker,
...
+     */
+    private Set<String> escapePaths(Set<String> paths) {
+        return paths.stream().map(s -> escapePath(s)).collect(Collectors.toSet());
+    }
+
+    /**
+     * Escape path for Windows system with boot@docker according to rules given here :
+     * https://github.com/boot2docker/boot2docker/blob/master/README.md#virtualbox-guest-additions
+     * @param path
+     * @return escaped set of path
```
evoevodin: It doesn't return escaped set of path, it returns escaped path.

evoevodin:
In general:

New code contains some formatting issues (javadoc), use this [formatter](https://github.com/codenvy/codenvy-dev-resources/tree/master/che-codestyle/src/main/resources).

What about injection of already escaped values?
I mean what if we need another stuff which is related to different operation system.
Maybe it would be better to solve this problem on binding level ? 

```java
if (SystemInfo.isWindows()) {
    install(new WindowsModule()).
} ...
```

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -115,6 +117,41 @@ public DockerInstanceProvider(DockerConnector docker,
...
 
+
+    /**
+     * Escape paths for Windows system with boot@docker according to rules given here :
+     * https://github.com/boot2docker/boot2docker/blob/master/README.md#virtualbox-guest-additions
+     * @param paths
+     * @return escaped set of path
+     */
+    private Set<String> escapePaths(Set<String> paths) {
+        return paths.stream().map(s -> escapePath(s)).collect(Collectors.toSet());
```
evoevodin: Use method reference

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
garagatyi: +1

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
garagatyi: I think it would be better to do this when we will need that. For now it's ok to me to put it here.

garagatyi: But,  please, move it below public methods.

evoevodin: Anyway it must be tested

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -115,6 +117,41 @@ public DockerInstanceProvider(DockerConnector docker,
...
+        String esc;
+        if (path.indexOf(":") == 1) { //check and replace only occurrence of ":" after disk label on Windows host (e.g. C:/)
+            // but keep other occurrences it can be marker for docker mount volumes (e.g. /path/dir/from/host:/name/of/dir/in/container                                               )
+            esc = path.replaceFirst(":", "").replace('\\', '/');
+        } else {
+            esc = path.replace('\\', '/');
+        }
+
+        if (!esc.startsWith("/")) {
+            esc = "/" + Character.toLowerCase(esc.charAt(0)) + esc.substring(1); //letter of disk marl must be lower case
```
garagatyi: What is `marl`?

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -115,6 +117,41 @@ public DockerInstanceProvider(DockerConnector docker,
...
+        } else {
+            esc = Character.toLowerCase(esc.charAt(0)) + esc.substring(1);//letter of disk marl must be lower case
```
garagatyi: `marl`?

vparfonov: typo, must be mark

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
vparfonov: ok

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
vparfonov: ok

PR #279, opened 2015-09-29
IDEX-3126: Provide path to the projects sources directory with metada…

garagatyi:
…ta of a dev machine

### DIFF
--- a/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
+++ b/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
@@ -47,8 +47,6 @@ import java.net.HttpURLConnection;
 import java.net.URL;
 import java.nio.file.Files;
 import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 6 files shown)


skabashnyuk:
LGTM

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -319,7 +321,8 @@ private Instance createInstance(String imageId,
                 volumes.addAll(devMachineSystemVolumes);
                 volumes.addAll(systemVolumesForMachine);
 
-                env = new String[] { apiEndpointEnvVariable };
+                env = new String[] { apiEndpointEnvVariable,
+                                     DockerInstanceMetadata.PROJECTS_ROOT_VARIABLE + "=" + PROJECTS_FOLDER_PATH };
```
evoevodin: Character literal <code>'='</code>  should be used rather than <code>"="</code> string literal in concatenation.

evoevodin:
:+1: looks nice

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -110,9 +111,19 @@ public DockerInstanceProvider(DockerConnector docker,
...
             machineContainerLabels.put("che:server:" + serverConf.getPort() + ":ref", serverConf.getRef());
+            devMachineContainerLabels.put("che:server:" + serverConf.getPort() + ":ref", serverConf.getRef());
             machineContainerLabels.put("che:server:" + serverConf.getPort() + ":protocol", serverConf.getProtocol());
+            devMachineContainerLabels.put("che:server:" + serverConf.getPort() + ":protocol", serverConf.getProtocol());
         }
+
+        this.systemVolumesForDevMachine.addAll(allMachinesSystemVolumes);
+        this.systemVolumesForDevMachine.addAll(systemVolumesForDevMachine);
+
+        commonEnvVariables = new String[0];
```
evoevodin: It looks like <code>commonEnvVariables</code> is a constant, so maybe it should be declared as <code>private static final</code> field ?

PR #282, opened 2015-09-29
IDEX-3085. Recognize project types after import raw project sources and make user possible to choose desired one.

RomanNikitenko:
Added ValueProviderFactory for each project type for recognizing project types after import raw project sources.


vparfonov:
ok

PR #295, opened 2015-10-02
IDEX-1752: appearing of Git Tab when sending a message to it

vinokurig:


plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/GitOutputPartPresenter.java
```diff
@@ -124,20 +159,33 @@ public void displayException(Exception e) {
...
     public void printError(String text) {
+        openTabIfClosed();
         view.printError(text);
         performPostOutputActions();
     }
 
+    /**
+     * Open Git Tab if closed
+     */
+    private void openTabIfClosed() {
```
sleshchenko: Maybe you can move this code into performPostOutputActions method

PR #301, opened 2015-10-06
IDEX-1999: Add button tab and buttons to git console

vinokurig:
removed button elements to che-core-ide-ui to use them in plugin-git and plugin-runner


plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/GitOutputPartViewImpl.java
```diff
@@ -136,4 +150,14 @@ public void scrollBottom() {
         scrollPanel.getElement().setScrollTop(scrollPanel.getElement().getScrollHeight());
     }
 
+    @NotNull
+    private void createButton(@NotNull SVGResource icon,
```
evoevodin: Align it. Please use [formatter](https://github.com/codenvy/codenvy-dev-resources/blob/master/che-codestyle/src/main/resources/che-codestyle-intellij.xml) for java source files.

vparfonov:
ok

skabashnyuk:
LGTM

PR #318, opened 2015-10-13
IDEX-1620: Disable buttons when got branch list

vinokurig:


skabashnyuk:
LGTM

sleshchenko:
Not OK for me. These changes produce set of bugs. For example, when user click on 'Create Branch' button and create new branch then after closing 'Create new branch' window all branches is disabled. You can use mechanism like https://github.com/codenvy/che-plugins/blob/IDEX-1620/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/remote/RemoteViewImpl.java#L218

So, view will check actuality of selection on setBranches. If selection is not actual view will send notice to presenter about it. Then presenter will disable buttons.

vinokurig:
fixed by adding a check that branch is not selected

plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/branch/BranchPresenter.java
```diff
@@ -309,6 +308,13 @@ protected void onFailure(Throwable exception) {
         }, null).show();
     }
 
+    @Override
+    public void onBranchUnselected() {
```
sleshchenko: I think you should reset value of current branch in this method.

sleshchenko:
other OK for me

skabashnyuk:
LGTM what about @vparfonov

plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/branch/BranchPresenter.java
```diff
@@ -104,11 +104,7 @@ public BranchPresenter(BranchView view,
     /** Show dialog. */
     public void showDialog() {
```
vparfonov: maybe need change method name, now it don't show dialog

plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/branch/BranchView.java
```diff
@@ -83,6 +86,14 @@
     /** Close dialog. */
     void close();
 
+    /**
```
vparfonov: Do we really need this method like public?

vparfonov:
ok

PR #323, opened 2015-10-19
IDEX-3138. Add tooltips on maven page. Fix NPE.

RomanNikitenko:
Add tooltips on maven page. 
Fix NPE.


plugin-java/che-plugin-java-ext-maven/src/main/java/org/eclipse/che/ide/extension/maven/client/wizard/MavenPagePresenter.java
```diff
@@ -33,6 +33,7 @@
 import java.util.List;
 import java.util.Map;
 
+import static com.google.common.base.MoreObjects.firstNonNull;
```
sleshchenko: I don't see guava-gwt dependency and inherit like <inherits name="com.google.common.base.Base"/>. Does it work by transitive dependency?

vparfonov: +1

vparfonov:
ok

PR #333, opened 2015-10-23
IDEX-3333 - adding README with drag&drop button

stour:
Adding a button for plugins that are not packaged by default with Che and that contain only one JAR.
(plugin-angularjs contains several JAR but is already packaged as che-plugin so button uses angularjs-assembly ZIP)


skabashnyuk:
LGTM

stour:
@benoitf just informed me that we are now able to use CURRENT as version in drag&drop button so I will update this PR accordingly

riuvshin:
@benoitf @stour this is awesome! thank you guys!

stour:
@skabashnyuk @vparfonov PR is now ready for merging
should I squash the two commits ?

benoitf:
I think it should be better to squash these two commits

stour:
done

PR #343, opened 2015-10-28
Revert "Revert "IDEX-2234 Removed autogenerating of ssh keys""

vinokurig:
This reverts commit that reverted removing autogenerating of ssh keys


skabashnyuk:
LGTM

PR #364, opened 2015-11-09
IDEX-3472 Use of Docker0 bridge name when we're on linux (bridge0 was not for Linux)

benoitf:
(see http://docs.docker.com/v1.8/articles/networking/) 
Change-Id: I6264bfcf93221a54f70079ff868cc4eccabcbd52


skabashnyuk:
+0 Have you checked on Linux?

benoitf:
well bridge0 was not the expected name when I've done the work.
Bridge name has to be 'docker0' so I'm pretty sure that today it doesn't work.(it use fallback ip for now)

garagatyi:
I think some guys already said that it is not working on Linux. On my machine bridge name is `docker0`. In devbox installation it is the same

garagatyi:
LGTM

garagatyi:
https://jira.codenvycorp.com/browse/IDEX-3472

PR #377, opened 2015-11-11
IDEX-3273: show an error message when update dependency is failed

svor:


plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/outputspanel/console/DefaultOutputConsole.java
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * Print message in the console.
+     *
+     * @param text
+     *         message which should be printed
+     */
+    public void printText(String text) {
+        view.print(text, false);
+        Scheduler.get().scheduleDeferred(new Scheduler.ScheduledCommand() {
```
azatsarynnyy: It would be nice if you left here a short explanation why you're using deferred command

svor: I've reworked it

vparfonov:
ok

PR #380, opened 2015-11-13
IDEX-3513: Disable OpenShift > Deploy > New Application in some case

AndrienkoAleksandr:
Disable OpenShift deploy for new application if a project is an OpenShift project type


plugin-openshift/che-plugin-openshift-ext-client/src/main/java/org/eclipse/che/ide/ext/openshift/client/deploy/_new/NewApplicationAction.java
```diff
@@ -49,7 +53,11 @@ public NewApplicationAction(final AnalyticsEventLogger eventLogger,
 
     @Override
     public void updateInPerspective(@NotNull ActionEvent event) {
+        event.getPresentation().setVisible(appContext.getCurrentProject() != null);
         event.getPresentation().setEnabled(authHandler.isLoggedIn() && appContext.getCurrentProject() != null);
+        event.getPresentation().setEnabled(appContext.getCurrentProject() != null
```
vzhukovskii: why you can't use method setVisibleAndEnable(boolean) and pass as argument result of preconditions?

PR #384, opened 2015-11-16
IDEX-3227: Add project modules to new project model

dimasnurenko:
@vparfonov @evidolob


skabashnyuk:
+0. Code is too large to review. You can proceed merging on your own responsibility.

PR #399, opened 2015-11-18
IDEX-2981: Add checkbox for recursive import in import panel

vinokurig:
We need git clone wirh --recursive to clone multi-module project with submodules.
Will be merged only in 3.x
@vparfonov @skabashnyuk


vparfonov:
ok

PR #424, opened 2015-11-25
IDEX-3396: Add ability to save workspace's snapshot

evoevodin:
Added component which checks if registry is available.
Bound `CreateSnapshotAction`

@skabashnyuk, @garagatyi, @akorneta, @sleshchenko please take a look


azatsarynnyy:
What's this: plugin-java/che-plugin-java-ext-jdt/che-jdt-ext-machine/index/2089348824.index

evoevodin:
:dizzy_face: :gun: I have not idea, i am going to remove it, thx

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerRegistryChecker.java
```diff
@@ -0,0 +1,73 @@
...
+    @Inject
+    @Named("docker.registry.auth.url")
+    private String registry;
+
+    /**
+     * Checks that registry is available and if it is not - logs warning message.
+     */
+    @PostConstruct
+    private void checkRegistryIsAvailable() throws IOException {
+        if (!probeUrl("http://" + registry) && !probeUrl("https://" + registry)) {
```
garagatyi: Registry can use https

skabashnyuk: can we ad protocol to url?

evoevodin: If you mean `docker.registry.auth.url` property then my answer is - no we can't, because 
docker will not be able to make push to the registry tagged as http://localhost:5000. In this case we can try to use several different properties.

skabashnyuk: Can you remove programmatically  protocol for push?

PR #428, opened 2015-11-27
CLDIDE-2633 Use repo instead of image id in docker machine impl

garagatyi:
Swarm has problems with scheduling containers if image id is used instead of repo. This PR fixes incorrect containers scheduling and allow system admin correlate docker image/container with workspace/machine

### DIFF
--- a/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceKey.java
+++ b/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceKey.java
@@ -63,6 +63,24 @@ public class DockerInstanceKey implements InstanceKey {
         return fields.get(REGISTRY);
     }
 
+    /**
+     * Returns full name of docker image.
+     *
+     * It consists of registry, userspace, repository name, tag.
+     * E.g. docker-registry.company.com:5000/userspace1/my-repository:some-tag
+     */
+    public String getFullName() {
+        final StringBuilder fullRepoId = new StringBuilder();
+        if (getRegistry() != null) {
+            fullRepoId.append(getRegistry()).append('/');
+        }
+        fullRepoId.append(getRepository());
+        if (getTag() != null) {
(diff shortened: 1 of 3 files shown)


plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceKey.java
```diff
@@ -63,6 +63,18 @@ public String getRegistry() {
         return fields.get(REGISTRY);
     }
 
+    public String getFullRepoId() {
```
evoevodin: I think in the case of this method it would be better to add some javadoc.

skabashnyuk: +

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -221,10 +221,44 @@ public Instance createInstance(Recipe recipe,
...
+        buildImage(dockerfile, creationLogsOutput, machineImageRepo);
+
+        return createInstance(machineContainerName,
+                              machineState,
+                              machineImageRepo,
+                              creationLogsOutput);
+    }
+
+    @Override
+    public Instance createInstance(InstanceKey instanceKey,
```
skabashnyuk: javadoc

skabashnyuk:
other ok

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
garagatyi: Method in interface has javadoc

evoevodin:
ok for me

PR #430, opened 2015-11-30
IDEX-3660: allow disabling force pull on docker build of machine

garagatyi:
@skabashnyuk @evoevodin Please review.

### DIFF
--- a/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
+++ b/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
@@ -219,12 +219,15 @@ public class DockerConnector {
      * @throws InterruptedException
      *         if build process was interrupted
      */
-    public String buildImage(String repository, ProgressMonitor progressMonitor, AuthConfigs authConfigs, File... files)
-            throws IOException, InterruptedException {
+    public String buildImage(String repository,
+                             ProgressMonitor progressMonitor,
+                             AuthConfigs authConfigs,
+                             boolean doForcePull,
+                             File... files) throws IOException, InterruptedException {
         final File tar = Files.createTempFile(null, ".tar").toFile();
         try {
             createTarArchive(tar, files);
-            return buildImage(repository, tar, progressMonitor, authConfigs);
+            return buildImage(repository, tar, progressMonitor, authConfigs, doForcePull);
         } finally {
             FileCleaner.addFile(tar);
         }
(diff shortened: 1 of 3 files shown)


plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -219,12 +219,15 @@ public Version getVersion() throws IOException {
      * @throws InterruptedException
      *         if build process was interrupted
      */
-    public String buildImage(String repository, ProgressMonitor progressMonitor, AuthConfigs authConfigs, File... files)
-            throws IOException, InterruptedException {
+    public String buildImage(String repository,
+                             ProgressMonitor progressMonitor,
+                             AuthConfigs authConfigs,
+                             boolean doForcePull,
```
benoitf: maybe if new params will be added middle-term, to introduce an object parameter containing the options.
Thus, signature won't be changed on many places each time a new param is added

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -820,9 +825,11 @@ protected String doBuildImage(String repository,
                                                             .method("POST")
                                                             .path("/build")
                                                             .query("rm", 1)
-                                                            .query("pull", 1)
```
evoevodin: As far as i know `true` may be used as query parameter value as well as 1, so you can just put `.query("pull", doForcePull);`

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
garagatyi: If so we should add config object to every method, no?

evoevodin: Yes i agree with @benoitf, it would be really nice to use objects in the case of `DockerConnector` methods.
But, as it requires a lot of changes you can add another method with a new argument and reuse this method in old one, which will save API compatibility.

garagatyi: After discussions we agreed to leave this PR with current style of parameters and create an issue for addition config objects to all methods of docker client. Is it OK for you guys?

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
garagatyi: yes, it works. Thanks for this hint!

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
benoitf: ok for me

evoevodin:
ok

skabashnyuk:
ok

PR #432, opened 2015-11-30
IDEX-3567: fix a bug with renaming packages

svor:


vparfonov:
ok for me, need wait on go from @evidolob

evidolob:
OK for me

PR #434, opened 2015-12-01
IDEX-3636: fix cursor positioning for custom suggestions in JS-editor

azatsarynnyy:
* fixed cursor positioning for custom suggestions in JS-editor (which was loaded from _proposals.json_);
* paramenter _overwrite_ in _proposals.json_ isn't use anymore: it will be always 'true';
* avoid using synchronous XMLHttpRequest - used Promises intsead;
* no need to reload the page for reloading the changes in _proposals.json_;
* added handling of errors of getting and parsing _proposals.json_.


vparfonov:
ok

PR #435, opened 2015-12-01
IDEX-3634 add Orion initial preferences

evidolob:
to enable braces completion feature


skabashnyuk:
That is the origin of this file?

evidolob:
It is downloaded from http://download.eclipse.org/orion/drops/R-10.0-201510301610/index.html

skabashnyuk:
ok

PR #438, opened 2015-12-03
Don't show popup when abnormal close code is gotten from websocket and reason is absent

RomanNikitenko:
Don't show popup when abnormal close code is gotten from websocket and reason is absent


vparfonov:
ok

PR #441, opened 2015-12-03
IDEX-3574 rename package of zeroCliboard

skabashnyuk:
To awoid problems with git checkout on case insensetive filesystem


benoitf:
usually packages are lowercases no ?
I would change it in the declaration instead of the import

skabashnyuk:
you are right. motivation answered in che-core pr

PR #442, opened 2015-12-04
IDEX-3744: disable docker machine state change detection

garagatyi:
Comment usage of container state change because it has bug.
Exec's OOM is treated as OOM of container.


skabashnyuk:
ок

benoitf:
do we need // FIXME in the commented lines ?

garagatyi:
Problem is that because of docker implementation of oom events we can't fix it without implementing oom detector on system level. I'm going to create issue in docker.
But I'll add Fixme comment

PR #444, opened 2015-12-04
IDEX-3658: Add delay when setting focus to message field in commit dialog

vinokurig:
Otherwise the focus on the text field wouldn't be settled
@vparfonov please review


plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/commit/CommitViewImpl.java
```diff
@@ -180,7 +181,12 @@ public void setEnableCommitButton(boolean enable) {
     /** {@inheritDoc} */
     @Override
     public void focusInMessageField() {
-        message.setFocus(true);
+        new Timer() {
```
sleshchenko: Try to do this with scheduler like https://github.com/codenvy/che-core/blob/88d75e053098e2455fa926df33916f88afdcb6c3/ide/che-core-ide-jseditor/src/main/java/org/eclipse/che/ide/jseditor/client/annotation/AnnotationModelImpl.java#L169-L175
To avoid unclear delay 300

vinokurig: Unfortunately it does not work in this place. It is a common practice to set such delay in our code.

vparfonov:
if it's only one way to fix it so ...
ok

PR #445, opened 2015-12-04
Fix invalid scope of the dependency

benoitf:
Change-Id: I8461926637975f3906bd47fd66838be79847735a


skabashnyuk:
ok

PR #447, opened 2015-12-07
IDEX-3744: fix treating non-main container process OOM as main proces…

garagatyi:
…s OOM


evoevodin:
I hope it works :+1:

skabashnyuk:
ok

skabashnyuk:
ok

PR #452, opened 2015-12-08
IDEX-3611 Add volume for providing external configuration for extensi…

sleshchenko:
…on tomcat

@skabashnyuk

### DIFF
--- a/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ext/DockerExtConfBindingProvider.java
+++ b/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ext/DockerExtConfBindingProvider.java
@@ -1,0 +1,46 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2015 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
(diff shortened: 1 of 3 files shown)


plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ext/DockerExtConfBindingProvider.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @author Sergii Leschenko
+ */
+public class DockerExtConfBindingProvider implements Provider<String> {
+    public static final String EXT_CHE_LOCAL_CONF_DIR = "/mnt/che/conf";
+
+    @Override
+    public String get() {
+        if (SystemInfo.isWindows()) {
+            return System.getProperty("user.home") + "\\AppData\\Local\\che\\ext-conf";
```
skabashnyuk: why Windows so special?

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ext/DockerExtConfBindingProvider.java
```diff
@@ -0,0 +1,44 @@
...
+        } else {
+            final String localConfDir = System.getenv(CheBootstrap.CHE_LOCAL_CONF_DIR);
+            String pathToExtConfDir = localConfDir == null ? "" : localConfDir + File.separator + "ext";
```
skabashnyuk: I have some doubts about this block. Need to think about alternatives

skabashnyuk:
ok

PR #453, opened 2015-12-08
IDEX-3631: Fix creating module in project which was cloned with keep directory option.

dimasnurenko:
@vparfonov


plugin-java/che-plugin-java-ext-maven/src/test/java/org/eclipse/che/ide/extension/maven/server/projecttype/handler/RemoveMavenModuleHandlerTest.java
```diff
@@ -128,8 +129,8 @@ public void cleanup() throws IllegalAccessException, NoSuchFieldException {
         f.set(null, new HttpJsonHelper.HttpJsonHelperImpl());
     }
 
-    @Test(expected = IllegalArgumentException.class)
-    public void shouldThrowExceptionWhenPomNotFound() throws Exception {
+    @Test
+    public void methodReturnedTheControlWhenPomNotFound() throws Exception {
```
sleshchenko: it would be better to use should in method name, no?

vparfonov:
ok

plugin-java/che-plugin-java-ext-maven/src/test/java/org/eclipse/che/ide/extension/maven/server/projecttype/handler/RemoveMavenModuleHandlerTest.java
dimasnurenko: Yes

PR #462, opened 2015-12-11
IDEX-3667: create new command wizard

olexii4:
@ashumilova  WDYT?


plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/command/CommandConfiguration.java
```diff
@@ -55,4 +55,50 @@ public CommandType getType() {
...
+                    return false;
+                }
+            }
+            return true;
+        }
+
+        @Override
+        public int hashCode() {
+            int hash = 7;
+            hash = hash * 31 + (name != null ? name.hashCode() : 0);
```
azatsarynnyy: Use *Objects.hashCode(name)* instead of *(name != null ? name.hashCode() : 0)*
And the same for other similar code

plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/command/CommandConfiguration.java
```diff
@@ -55,4 +55,50 @@ public CommandType getType() {
     /** Returns command line to execute in machine. */
     @NotNull
     public abstract String toCommandLine();
+    
+        @Override
+        public boolean equals(Object o) {
```
azatsarynnyy: this method may be really simplified with Objects.equals() method

plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/MachineLocalizationConstant.java
```diff
@@ -26,9 +26,15 @@
...
 
+    @Key("button.save")
+    String saveButton();
+
     @Key("button.cancel")
     String cancelButton();
 
+    @Key("button.close")
+    String closeButton();
+
```
ashumilova: all this button names can be used from org.eclipse.che.ide.CoreLocalizationConstant
no need to duplicate these

azatsarynnyy:
+1

PR #463, opened 2015-12-11
IDEX-2982. Change Che pom.xml parent values

RomanNikitenko:
Change Che pom.xml parent values


skabashnyuk:
ok. @vparfonov ?

plugin-sdk/pom.xml
```diff
@@ -19,7 +19,7 @@
         <version>4.0.0-M5-SNAPSHOT</version>
         <relativePath>../pom.xml</relativePath>
     </parent>
-    <artifactId>che-plugin-sdk-parent</artifactId>
+    <artifactId>che-plugin-ide-parent</artifactId>
     <packaging>pom</packaging>
     <name>Che Plugin :: SDK :: Parent</name>
```
vzhukovskii: Maybe we should change also name tag according to artifact name?

vparfonov:
ok

PR #464, opened 2015-12-11
Idex 3437 add possibility to trigger docker API on host from container

garagatyi:


rainf0x:
Ok

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -187,6 +151,7 @@ public Version getVersion() throws IOException {
      * @return list of docker images
      * @throws IOException
      */
+    @SuppressWarnings("unused")
```
evoevodin: I don't think that it is the place for this suppression, as it is API method. I think you should suppress 'unused' warning in the situations when there is no way to use it. If it is unused in your code in doesn't mean that it is unused at all + what is the purpose of the `unused` API method? what if someone will use it, should he remove suppression? The nice [place](https://github.com/codenvy/che-core/blob/2aa05905269a75d358ac282260d927bdca232d77/platform-api/che-core-api-core/src/test/java/org/eclipse/che/api/core/rest/DefaultHttpJsonRequestTest.java#L72-L73) to use `@SuppressWarnings("unused")`

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
```diff
@@ -102,13 +129,53 @@ public DockerConnectorConfiguration(URI dockerDaemonUri,
...
+        return UNIX_SOCKET_SCHEME.equals(uri.getScheme());
+    }
+
+    public String getDockerHost() {
+        if (isUnixSocketUri(dockerDaemonUri)) {
+            return "localhost";
+        } else {
+            return getDockerHostIp();
+        }
+    }
```
mshaposhnik: why do you hate ternary op ?

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
```diff
@@ -102,13 +129,53 @@ public DockerConnectorConfiguration(URI dockerDaemonUri,
         this.networkFinder = networkFinder;
     }
 
-    public static String getExpectedLocalHost() {
-        return SystemInfo.isLinux() ? "localhost" : dockerDaemonUri().getHost();
+    public static boolean isUnixSocketUri(URI uri) {
```
evoevodin: What if `uri` is null ?
Don't you think that it would be better to define this sort of contract here?

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
garagatyi: Agree

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
```diff
@@ -102,13 +129,53 @@ public DockerConnectorConfiguration(URI dockerDaemonUri,
...
+    public DockerCertificates getDockerCertificates() {
+        if (dockerCertificatesDirectoryPath == null || !getDockerDaemonUri().getScheme().equals("https")) {
+            return null;
+        }
+        final File dockerCertificatesDirectory = new File(dockerCertificatesDirectoryPath);
+        return dockerCertificatesDirectory.isDirectory() ? DockerCertificates.loadFromDirectory(dockerCertificatesDirectory) : null;
+    }
+
+    /**
+     * For testing purposes
```
evoevodin: Then it shouldn't be public, or maybe you forgot to remove either comment or method?

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
```diff
@@ -203,23 +264,18 @@ protected String getDockerHostIp(final boolean isLinux, @NotNull final Map<Strin
...
-     *         if System is running on Linux
      * @param env
      *         should contain System environment
      * @return local path of the docker certificates
      */
-    protected static String dockerMachineCertsDirectoryPath(boolean isLinux, @NotNull Map<String, String> env) {
-        if (isLinux) {
-            return null;
-        }
+    protected static String dockerMachineCertsDirectoryPath(@NotNull Map<String, String> env) {
```
evoevodin: If it is required not to be `null` may be you should add appropriate check e.g `Objects.requireNonNull(env);`

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
garagatyi: Just played with code and when didn't see it when was finishing code cleanup.

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
garagatyi: @NotNull will be ok for you?

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ext/DockerTerminalModule.java
```diff
@@ -40,10 +45,29 @@ protected void configure() {
...
-            volumesMultibinder.addBinding().toProvider(TerminalServerBindingProviderWinOS.class);
-        } else {
-            volumesMultibinder.addBinding().toProvider(TerminalServerBindingProviderUnix.class);
+        volumesMultibinder.addBinding().toProvider(TerminalVolumeProvider.class).in(Singleton.class);
+    }
+
+    /**
+     * Provides volumes configuration of machine for terminal
+     *
+     * On Windows MUST be locate in "user.home" directory in case limitation windows+docker.
```
evoevodin: In this case `On Windows MUST...` will be in the same line as `for terminal` is, what is worth is that it will be the only sentence, i see that you moved this code from different place, but anyway i think you should fix it, as it is broken

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
garagatyi: I'll remove it, thank you.

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
evoevodin: I think you should use `@NotNull` + `Objects.requireNonNull(uri, "Required non-null uri")` and write javadoc that explains NotNull contract, e.g. defines `@throws NullPointerException` with appropriate description.

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
garagatyi: ok

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ext/DockerTerminalModule.java
garagatyi: Thank you. I'll fix it

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
garagatyi: I think that @NotNull make it clear enough without javadocs, no?

evoevodin:
Generally it looks good! my +1
I am worrying about size of the docker API client, it grows and grows, which is generally good, the only thing i am worrying about is that it is kind of vertical growing. I think we should start thinking about sort of api modularisation which will help us to grow horizontally.

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
evoevodin: It does the half part of the contract, as it doesn't say about method reaction on null parameter.
Should i expect `NullPointerException`, `IllegalArgumentException` or maybe `BadRequestException`, from the `@NotNull` annotation it is not clear + i don't think that javadoc generators respect this annotation.

garagatyi:
I'm not happy with 1K+ lines classes either.

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
garagatyi: ok, I'll add it to javadocs

garagatyi:
Here is example of docker client library with separation of methods in different classes https://github.com/docker-java/docker-java

benoitf:
you may think about decoupling each method of connector into dedicated classes and replace invocation code logic which is quite duplicated
try (DockerConnection connection = connectionFactory.openConnection(dockerDaemonUri)
                                                            ......) {
            final DockerResponse response = connection.request();
            final int status = response.getStatus();
            if (... != status) {

with helper, more injection of smaller objects

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ext/DockerExtServerModule.java
```diff
@@ -37,16 +45,65 @@ protected void configure() {
...
 
         DockerExtConfBindingProvider extConfBindingProvider = new DockerExtConfBindingProvider();
         if (extConfBindingProvider.get() != null) {
             volumesMultibinder.addBinding().toProvider(extConfBindingProvider);
         }
     }
+
+    /** Add env variable to docker dev-machine with url of Che API */
+    private static class ApiEndpointEnvVariableProvider implements Provider<String> {
+        @Inject
```
skabashnyuk: Maybe this class is worth of separate file?

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ext/DockerExtServerModule.java
```diff
@@ -37,16 +45,65 @@ protected void configure() {
...
+
+        @Override
+        public String get() {
+            return DockerInstanceMetadata.API_ENDPOINT_URL_VARIABLE + '=' + apiEndpoint;
+        }
+    }
+
+    /** Add env variable to docker dev-machine with path to root folder of projects */
+    private static class ProjectsRootEnvVariableProvider implements Provider<String> {
+        @Inject
```
skabashnyuk: Maybe this class is worth of separate file?

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ext/DockerExtServerModule.java
```diff
@@ -37,16 +45,65 @@ protected void configure() {
...
+        }
+    }
+
+    /**
+     * Reads path to extensions server archive to mount it to docker machine
+     *
+     * <p>On Windows hosts MUST be locate in "user.home" directory in case limitation windows+docker.
+     */
+    private static class ExtServerVolumeProvider implements Provider<String> {
+        @Inject
```
skabashnyuk: Maybe this class is worth of separate file?

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ext/DockerTerminalModule.java
```diff
@@ -40,10 +46,27 @@ protected void configure() {
...
+    }
+
+    /**
+     * Provides volumes configuration of machine for terminal
+     *
+     * <p>On Windows MUST be locate in "user.home" directory in case limitation windows+docker.
+     *
+     */
+    private static class TerminalVolumeProvider implements Provider<String> {
+        @Inject
```
skabashnyuk: Maybe this class is worth of separate file?

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ext/LocalStorageModule.java
```diff
@@ -27,4 +31,16 @@ protected void configure() {
                 Multibinder.newSetBinder(binder(), String.class, Names.named("machine.docker.dev_machine.machine_volumes"));
         volumesMultibinder.addBinding().toProvider(LocalStorageDockerVolumePathProvider.class);
     }
+
+    /** Provides volumes configuration of machine for local storage */
+    private static class LocalStorageDockerVolumePathProvider implements Provider<String> {
```
skabashnyuk: Maybe this class is worth of separate file?

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalDockerModule.java
```diff
@@ -54,17 +56,63 @@ protected void configure() {
...
+    }
+
+    /**
+     * Provide path to the project folder on hosted machine
+     *
+     * <p>On Unix managed by vfs.local.fs_root_dir property.<br>
+     * On Windows MUST be locate in "user.home" directory in case limitation windows+docker
+     */
+    private static class HostProjectFolderProvider implements Provider<String> {
+        @Inject
```
skabashnyuk: Maybe this class is worth of separate file?

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalDockerModule.java
```diff
@@ -54,17 +56,63 @@ protected void configure() {
...
+            if (SystemInfo.isWindows()) {
+                return System.getProperty("user.home") + "\\che\\projects";
+            } else {
+                return projectsFolder;
+            }
+        }
+    }
+
+    /** Provides DOCKER_HOST env variable for the sake of access to docker API within docker container */
+    private static class DockerApiHostEnvVariableProvider implements Provider<String> {
```
skabashnyuk: Maybe this class is worth of separate file?

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
```diff
@@ -102,13 +130,48 @@ public DockerConnectorConfiguration(URI dockerDaemonUri,
...
+     * Tests URI to check if it is URI of UNIX socket
+     *
+     * @throws NullPointerException if argument is null
+     */
+    public static boolean isUnixSocketUri(@NotNull  URI uri) {
+        Objects.requireNonNull(uri, "Required non-null uri");
+        return UNIX_SOCKET_SCHEME.equals(uri.getScheme());
+    }
+
+    public String getDockerHost() {
```
skabashnyuk: what is it and where it's used?

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
```diff
@@ -102,13 +130,48 @@ public DockerConnectorConfiguration(URI dockerDaemonUri,
...
+     *
+     * @return docker host IP address
+     * @see <a href="https://docs.docker.com/articles/networking/">Docker Networking</a>
+     */
+    public String getDockerHostIp() {
+        return getDockerHostIp(SystemInfo.isLinux(), System.getenv());
+    }
+
+    public URI getDockerDaemonUri() {
+        return dockerDaemonUri;
```
skabashnyuk: what is it and where it's used?
javadoc?

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
```diff
@@ -102,13 +130,48 @@ public DockerConnectorConfiguration(URI dockerDaemonUri,
...
+    }
+
+    public InitialAuthConfig getAuthConfigs() {
+        return authConfigs;
```
skabashnyuk: javadoc

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
```diff
@@ -102,13 +130,48 @@ public DockerConnectorConfiguration(URI dockerDaemonUri,
...
+    }
+
+    public DockerCertificates getDockerCertificates() {
+        if (dockerCertificatesDirectoryPath == null || !getDockerDaemonUri().getScheme().equals("https")) {
```
skabashnyuk: javadoc

plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
garagatyi: It is used in LocalDockerNode. I'll add javadocs.

skabashnyuk:
ok

PR #466, opened 2015-12-14
IDEX-3817 Small fixes after review for console panel UI

RomanNikitenko:
#1 - Fix displaying of Icon for the Consoles panel
#2 - Fix naming of terminals: Terminal-## instead of Terminal (##)
#3 - Number for terminals start with "Terminal" then the second one "Terminal-2" and then we go incrementaly
#4 - Add title "Consoles" for the panel
#5 - Change displaying of 'Add new terminal' icon
#6 - Remove bold for "command:" in the output view


vparfonov:
ok

PR #468, opened 2015-12-15
IDEX-3818. Add ability to show output of executing command for several projects

RomanNikitenko:
Show output of executing command in new panel if current command isn't stopped/finished
Show output of executing command in the same panel if current command is stopped/finished


vparfonov:
ok

PR #471, opened 2015-12-15
IDEX-3437: fix binding of named variables

garagatyi:


plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineModule.java
```diff
@@ -0,0 +1,46 @@
...
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.plugin.docker.machine;
+
+import com.google.inject.AbstractModule;
+import com.google.inject.multibindings.Multibinder;
+import com.google.inject.name.Names;
+
+/**
+ *
```
skabashnyuk: Can you describe what this module contains?

garagatyi: Yep, I forgot. 10x

garagatyi: done

PR #475, opened 2015-12-16
IDEX-3736: allow setting custom docker containers host

garagatyi:


plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineFactory.java
```diff
@@ -39,4 +40,7 @@ Instance createInstance(@Assisted MachineState machineState,
 
     DockerNode createNode(@Assisted("workspace") String workspaceId,
                           @Assisted("container") String containerId) throws MachineException;
+
+    DockerInstanceMetadata createMetadata(@Assisted ContainerInfo containerInfo,
```
vzhukovskii: Javadoc

vzhukovskii: What if I'll pass null as parameters?

garagatyi: Assisted injection doesn't allow pass null, Guice will throw exception.

skabashnyuk:
ок

PR #477, opened 2015-12-16
IDEX-3632. Don't duplicate message about starting machine for dev macnine

RomanNikitenko:
We show the loader for dev machine when this one is starting 
So we shouldn't duplicate message in machine panel


vparfonov:
ok

PR #489, opened 2015-12-21
IDEX-3684: add name to machine process

garagatyi:


skabashnyuk:
Did you do something else except then "add name"?

garagatyi:
I changed design of docker implementation because in current implementation there is no place to save name.
I refactor code that returns container's metadata. Now it is easier to support different implementations of MachineMetadata.
I removed DockerRuntimeException because it works in a different way to the common behaviour of the system.

skabashnyuk:
ok

PR #494, opened 2015-12-22
IDEX-3775: Fix situation when push command return 200 ok but in fact …

akorneta:
…produce error


garagatyi:
ok

skabashnyuk:
ok. @evoevodin ?

evoevodin:
:+1:

PR #496, opened 2015-12-23
IDEX-3830. Fix behavior of terminal after re-focus

RomanNikitenko:
Fix behavior of terminal after re-focus
Set focus on terminal after resize
Fix NPE at resizing before terminal is connect
Add ability to select text in processes panel


vparfonov:
ok

skabashnyuk:
ok

PR #516, opened 2015-12-30
IDEX-3920 Adopt to new SSH API

sleshchenko:


skabashnyuk:
ok

plugin-git/che-plugin-git-ssh-client/src/main/java/org/eclipse/che/ide/ext/ssh/client/GitSshKeyUploaderRegistry.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Map;
+
+/**
+ * Holds available ssh key uploaders for git
+ *
+ * @author Sergii Leschenko
+ */
+@Singleton
+public class GitSshKeyUploaderRegistry {
+    private final Map<String, SshKeyUploader> sshKeyUploaders = new HashMap<>();
```
evoevodin: This registry is not thread-safe

PR #519, opened 2016-01-04
IDEX-3929. Restore processes panes after page refreshing

RomanNikitenko:
Restore processes panes after page refreshing

### DIFF
--- a/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstance.java
+++ b/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstance.java
@@ -13,6 +13,7 @@ package org.eclipse.che.plugin.docker.machine;
 import com.google.inject.assistedinject.Assisted;
 
 import org.eclipse.che.api.core.NotFoundException;
+import org.eclipse.che.api.core.model.machine.Command;
 import org.eclipse.che.api.core.model.machine.MachineMetadata;
 import org.eclipse.che.api.core.model.machine.MachineState;
 import org.eclipse.che.api.core.util.LineConsumer;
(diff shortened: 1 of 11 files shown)


vparfonov:
ok
@garagatyi

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstance.java
```diff
@@ -169,11 +170,13 @@ public InstanceProcess getProcess(final int pid) throws NotFoundException, Machi
     }
 
     @Override
-    public InstanceProcess createProcess(String commandName, String commandLine) throws MachineException {
+    public InstanceProcess createProcess(Command command, String outputChannel) throws MachineException {
         final Integer pid = pidSequence.getAndIncrement();
         final InstanceProcess process = dockerMachineFactory.createProcess(container,
```
garagatyi: use Command instead of separate parameters

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineFactory.java
```diff
@@ -37,6 +37,8 @@
     InstanceProcess createProcess(@Assisted("container") String container,
                                   @Assisted("commandName") String commandName,
```
garagatyi: use Command instead of separate parameters

plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstance.java
vparfonov: +1

garagatyi:
ok

garagatyi:
Please squash before merge

PR #522, opened 2016-01-04
IDEX-3924: Display preview URL in console panel

vzhukovskii:
@slemeur @vparfonov @ashumilova @skabashnyuk

### DIFF
--- a/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/command/valueproviders/CurrentProjectPathProvider.java
+++ b/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/command/valueproviders/CurrentProjectPathProvider.java
@@ -30,6 +30,7 @@ import javax.validation.constraints.NotNull;
 
 /**
  * Provides current project's path.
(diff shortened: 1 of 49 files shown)


plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/command/valueproviders/CurrentProjectPathProvider.java
```diff
@@ -41,7 +42,7 @@
                                                    ProjectReadyHandler,
                                                    CurrentProjectChangedHandler {
 
-    private static final String KEY = "${current.project.path}";
+    public static final String KEY = "${current.project.path}";
```
azatsarynnyy: Why you changed it to ```public``` while class has an appropriate public ```getKey()``` method?

vzhukovskii: Because I had to place this property name in other place, such as Preview Url template.

vzhukovskii:
Updated state:

PR #524, opened 2016-01-05
IDEX-4010 Add injection of user's ssh public keys

sleshchenko:
@skabashnyuk @garagatyi


garagatyi:
@vparfonov @evidolob WDYT?

garagatyi:
Maybe it would be better to replace docker exec with machine process, [e.g.](https://github.com/codenvy/che-core/blob/master/platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/WsAgentLauncherImpl.java) ? WDYT

sleshchenko:
I think In my case docker exec is more simple.

garagatyi:
ok
@skabashnyuk WDYT?

skabashnyuk:
ok

vparfonov:
ok

PR #525, opened 2016-01-05
Fix dependency

benoitf:
Change-Id: I7fc7f4f8fdb2352f189448cfe9aef3ca4285b601


garagatyi:
Good catch. OK

skabashnyuk:
ok

PR #532, opened 2016-01-07
Fix dependency scope

benoitf:
Change-Id: Ibd5bf304aa71f5d1318a0b6effb165e26dd350a0


skabashnyuk:
ok

PR #535, opened 2016-01-12
IDEX-4042 Improve slider in the console

vitaliy-guliy:


plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/processes/ConsolesPanelViewImpl.java
```diff
@@ -169,6 +175,66 @@ public void onKeyboard(KeyboardEvent event) {
...
+
+    /**
+     * Tunes splitter. Makes it wider and adds double border to seem rich.
+     *
+     * @param el
+     *         element to tune
+     */
+    private void tuneSplitter(Element el) {
+        /** Add Z-Index to move the splitter on the top and make content visible */
+        el.getParentElement().getStyle().setProperty("zIndex", "1000");
```
vzhukovskii: Maybe it would be better to move this into .css file?

PR #537, opened 2016-01-12
IDEX-3896. Display the panel for managing recipes regardless of whether a project exists in workspace

RomanNikitenko:
Display the panel for managing recipes regardless of whether a project exists in workspace
Fix bug related with incorrect displaying recipes after switch between perspectives
Fix displaying list of recipes(css).


vparfonov:
ok

PR #541, opened 2016-01-13
Fetch machine host name from server configuration instead of client W…

vzhukovskii:
…indow.Location

@vparfonov this need to proper detection of current machine host name/ip address

### DIFF
--- a/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/command/valueproviders/MachineHostNameProvider.java
+++ b/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/command/valueproviders/MachineHostNameProvider.java
@@ -10,20 +10,45 @@
  *******************************************************************************/
 package org.eclipse.che.ide.extension.machine.client.command.valueproviders;
 
-import com.google.gwt.user.client.Window;
+import com.google.inject.Inject;
 import com.google.inject.Singleton;
+import com.google.web.bindery.event.shared.EventBus;
+
+import org.eclipse.che.api.machine.gwt.client.MachineServiceClient;
+import org.eclipse.che.api.machine.shared.dto.MachineDto;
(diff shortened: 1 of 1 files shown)


garagatyi:
Why you inject it for dev machines only?

plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/command/valueproviders/MachineHostNameProvider.java
```diff
@@ -32,6 +57,59 @@ public String getKey() {
...
+        }).catchError(new Operation<PromiseError>() {
+            @Override
+            public void apply(PromiseError arg) throws OperationException {
+                value = "";
+            }
+        });
+    }
+
+    private String getHostName(MachineDto machine) {
+        for (ServerDto server : machine.getMetadata().getServers().values()) {
```
garagatyi: Is it possible that this method will be called when machine is in state STARTING?

vzhukovskii: org.eclipse.che.api.machine.shared.dto.event.MachineStatusEvent.EventType

Where I can see smth about starting event?

vzhukovskii:
> Why you inject it for dev machines only?
Because at this moment we have only one machine.

garagatyi:
I'm sure user can create several machines.

plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/command/valueproviders/MachineHostNameProvider.java
garagatyi: Find model interfaces instead of DTOs

vparfonov:
need to rework

PR #544, opened 2016-01-14
IDEX-4039: Rename Find tab to Usages

vzhukovskii:
@vparfonov @ashumilova 
this need to visual identify tabs if we use find text and usages in the same time


ashumilova:
+1 :)

vzhukovskii:
Done

PR #545, opened 2016-01-14
IDEX-4025: Change color of the stop button in console

vzhukovskii:
Result on Dark Theme

Result on Light Theme


slemeur:
Fine for me

PR #550, opened 2016-01-15
Notice user that projects must be built before launching GWT SDM for Che command

azatsarynnyy:
Several guys have asked me why they can't launch *GWT SDM for Che* command for just cloned projects. This is because their projects are not built yet.
Such warning may prevent user to be confused why *GWT SDM for Che* command doesn't work when projects aren't built yet.
@vparfonov


vparfonov:
ok

PR #556, opened 2016-01-15
IDEX-4084 Midnight Commander is not correctly displayed in terminal under OSX

vitaliy-guliy:


plugin-machine/che-plugin-machine-ext-client/src/main/resources/org/eclipse/che/ide/extension/machine/client/machine.css
```diff
@@ -23,12 +23,15 @@
 @external .terminal, .terminal-cursor;
 
 .terminal {
-    border-color: outputBackgroundColor;
+    /*border-color: outputBackgroundColor;*/
```
vzhukovskii: do we really need this commented code?

vitaliy-guliy: IDE will not work without this comment

skabashnyuk:
awesome  @vparfonov  wdyt?

PR #560, opened 2016-01-15
IDEX-3964: Resources folders shouldn't be displayed as java packages …

vzhukovskii:
…but as simple folders

@vparfonov


vparfonov:
ok

PR #567, opened 2016-01-16
IDEX-4041: Memory leak in static maps in java debugger

tolusha:


tolusha:
@vparfonov 
@skabashnyuk

tolusha:
The idea of the patch:
since we caught disconnect event no need to perform disconnect again. Exception occurs and Debugger instance remains in map.

vparfonov:
+1

skabashnyuk:
ok

PR #569, opened 2016-01-16
IDEX-4068: add highlighting of 'this' keyword, 'super' keyword and annotations in Java-files

azatsarynnyy:
@vparfonov


vparfonov:
ok

PR #572, opened 2016-01-18
IDEX-4106: fix problem with founding project by full resource path.

vparfonov:
We get all project from workspace found project by path of resources by select longest path of project that match current resource path


skabashnyuk:
ok

PR #577, opened 2016-01-19
Code clean-up: remove GWT project type. It was created on early stage of 4.0 when recipe was defined in PT

azatsarynnyy:
Removed GWT project type. It was created on early stage of 4.0 when recipe was defined in PT.
Now it's unnecessary.
@vparfonov


skabashnyuk:
ok for me. @vparfonov ?

vparfonov:
ok

skabashnyuk:
ok

PR #578, opened 2016-01-19
Fix incorrect state stop process button

RomanNikitenko:
Fix incorrect state stop process button
Fix duplicate set active part


skabashnyuk:
ok for me. @vparfonov ?

vparfonov:
ok

PR #588, opened 2016-01-21
IDEX-2389: Add getProjectExplorerSelection method in SelectionAgent

vinokurig:
When performing git commit with "add selected files" check-box selected and console part is active, NPE appears because the getSelection method in Selection agent returned a selection of console part, but we need a selection of project explorer
@vparfonov @skabashnyuk  please review


skabashnyuk:
ok for me. @vparfonov ?

azatsarynnyy:
see https://github.com/codenvy/che-core/pull/660#issuecomment-173586576

vinokurig:
Used Project explorer presenter instead of Selection agent

vparfonov:
ok

PR #596, opened 2016-01-25
CLDIDE-2682: Add git checkout action

akorneta:
Note if the branch to checkout is no exist it will create new one from startPoint/default and checkout on it.


vparfonov:
ok, after release 4.0-beta13

skabashnyuk:
ok

PR #597, opened 2016-01-25
IDEX-3955:Use handler for catching event then content inject to Editor Widget

vparfonov:
Set cursor position only then content already added to editor but in some case we need a little timeout in OpenDeclarationFinder (set it to 100 ms)

@evidolob @azatsarynnyy


azatsarynnyy:
+1

PR #603, opened 2016-01-28
Disable fail build by findbugs

garagatyi:
Disable findbugs build fail in modiles that fails if
enable it by default

@skabashnyuk @vparfonov


skabashnyuk:
ok

PR #610, opened 2016-02-01
IDEX-3940:Add to index new files while renaming refactoring

vinokurig:
@skabashnyuk @vparfonov please review

### DIFF
--- a/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/AddToIndexHandler.java
+++ b/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/AddToIndexHandler.java
@@ -16,12 +16,12 @@ import org.eclipse.che.api.git.gwt.client.GitServiceClient;
 import org.eclipse.che.api.git.shared.Status;
 import org.eclipse.che.api.workspace.shared.dto.ProjectConfigDto;
 import org.eclipse.che.ide.api.app.AppContext;
-import org.eclipse.che.ide.event.CreateFileNodeEvent;
-import org.eclipse.che.ide.event.CreateFileNodeEventHandler;
-import org.eclipse.che.ide.event.DeleteNodesEvent;
-import org.eclipse.che.ide.event.DeleteNodesEventHandler;
-import org.eclipse.che.ide.event.RenameNodeEvent;
-import org.eclipse.che.ide.event.RenameNodeEventHandler;
+import org.eclipse.che.ide.event.FileNodeCreatedEvent;
+import org.eclipse.che.ide.event.FileNodeCreatedEventHandler;
+import org.eclipse.che.ide.event.NodesDeletedEvent;
+import org.eclipse.che.ide.event.NodesDeletedEventHandler;
+import org.eclipse.che.ide.event.NodeRenamedEvent;
+import org.eclipse.che.ide.event.NodeRenamedEventHandler;
 import org.eclipse.che.ide.api.notification.Notification;
 import org.eclipse.che.ide.api.notification.NotificationManager;
 import org.eclipse.che.ide.api.notification.StatusNotification;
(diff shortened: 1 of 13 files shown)


plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/AddToIndexManager.java
```diff
@@ -0,0 +1,318 @@
...
+import java.util.List;
+import java.util.Map;
+
+import static org.eclipse.che.ide.api.notification.StatusNotification.Status.FAIL;
+import static org.eclipse.che.ide.api.notification.StatusNotification.Status.SUCCESS;
+
+/**
+ * @author Igor Vinokur
+ */
+public class AddToIndexManager implements CreateFileNodeEventHandler, RefactoringProvidedEventHandler, DeleteNodesEventHandler,
```
azatsarynnyy: I would say that this is more **AddToIndexListener** or **AddToIndexHandler** than **AddToIndexManager**.
WDYT?

vinokurig: Renamed to **AddToIndexManager**

vzhukovskii: @azatsarynnyy may be it would be better to rename this class simply to `IndexManager` and create several usefull methods to control index state? I suppose that we will have methods like `removeFromIndex` also in future...

vparfonov:
not goof have dependency in git plugin on java plugin

plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/AddToIndexHandler.java
```diff
@@ -0,0 +1,318 @@
...
+        } else {
+            return "";
+        }
+    }
+
+    private boolean isUnderGit() {
+        //TODO rework checking for git mixin when IDEX-3725 will be fixed
+        final ProjectConfigDto project = appContext.getCurrentProject().getRootProject();
+        Map<String, List<String>> attributes = project.getAttributes();
+        return attributes.containsKey("vcs.provider.name") && attributes.get("vcs.provider.name").contains("git");
```
azatsarynnyy: It would be better to move [these](https://github.com/codenvy/che-core/blob/project_api_4/platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitProjectType.java#L24-L26) constants to the shared class (see example [here](https://github.com/codenvy/che-core/blob/project_api_4/platform-api/che-core-api-ssh/src/main/java/org/eclipse/che/api/ssh/shared/Constants.java)) and use it there.

plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/AddToIndexHandler.java
```diff
@@ -0,0 +1,318 @@
...
+        CancelCallback cancelCallback = new CancelCallback() {
+            @Override
+            public void cancelled() {
+                //Do nothing
+            }
+        };
+        dialogFactory.createConfirmDialog(locale.newFileAddToIndexDialogTitle(),
+                                          locale.newFileAddToIndexDialogText(fileName),
+                                          YES_BUTTON,
+                                          NO_BUTTON,
```
azatsarynnyy: move buttons names to localization constants

plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/AddToIndexHandler.java
vinokurig: used constant from GitProjectType class

plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/AddToIndexHandler.java
vinokurig: done

vinokurig:
Need to rework according to new vfs events that will be implemented in IDEX-4055

PR #614, opened 2016-02-04
IIU-384: fix always opened debugger tab after refresh page

mmorhun:
Fix problem with debugger tab, which opens after each refresh even if user hide it. Now it opens when connect to debugger or stop on breakpoint.


mmorhun:
@vparfonov please review

mmorhun:
@vparfonov please check

tolusha:
ok

vparfonov:
ok

PR #616, opened 2016-02-05
IDEX-4215 Updates icons so they are all compatible with Eclipse Licensing

vitaliy-guliy:


plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/action/CompareWithBranchAction.java
```diff
@@ -33,7 +33,7 @@ public CompareWithBranchAction(BranchListPresenter presenter,
                                    AppContext appContext,
                                    GitLocalizationConstant locale,
                                    ProjectExplorerPresenter projectExplorer) {
-        super(locale.compareWithBranchTitle(), locale.compareWithBranchTitle(), null, appContext, projectExplorer);
+        super(locale.compareWithBranchTitle(), locale.compareWithBranchTitle(), (String)null, appContext, projectExplorer);
```
vparfonov: (String)null for what?

vitaliy-guliy: An extra constructor added to base org.eclipse.che.ide.ext.git.client.action.GitAction

PR #617, opened 2016-02-05
IDEX-3531: Item is removed from project tree after remove from index

dimasnurenko:
@vparfonov


plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/remove/RemoveFromIndexPresenter.java
```diff
@@ -194,7 +194,7 @@ protected void onFailure(Throwable exception) {
     protected ResourceBasedNode<?> getResourceBasedNode() {
         List<?> selection = projectExplorer.getSelection().getAllElements();
         //we should be sure that user selected single element to work with it
-        if (selection != null && selection.size() > 1) {
+        if (selection != null && selection.size() >= 1) {
```
azatsarynnyy: what about `!selection.isEmpty()` instead of `selection.size() >= 1`?

vparfonov:
ok

PR #623, opened 2016-02-09
Remove unnecessary -n cp argument

eivantsov:
There are two reasons here:

1. -n option is really unnecessary since we remove /home/user/che and recreate it here - https://github.com/codenvy/che/blob/master/assembly-ide-war/src/main/java/org/eclipse/che/api/deploy/ApiModule.java#L122 thus there cannot be a situation when a terminal folder exists in /home/user/che

2. Alpine and busybox do not recognize -n with cp. This it is impossible now to use alpine and busybox based images.


garagatyi:
ok for me.
@akorneta WDYT?

skabashnyuk:
ok

